PR #88, opened 2024-08-26
feat: add visionos support

okwasniewski:
This PR adds visionOS support.

As I still can't get the build to work locally I decided to try to build dawn library on the CI


package/scripts/build/dawn.ts
```diff
@@ -24,6 +24,8 @@ const PLATFORM_MAP: Record<string, string> = {
   arm64_iphoneos: "OS64",
   arm64_iphonesimulator: "SIMULATORARM64",
   x86_64_iphonesimulator: "SIMULATOR64",
+  arm64_xros: "VISIONOS",
+  arm64_xrsimulator: "SIMULATOR_VISIONOS",
```
wcandillon, line 28: there is no x86_64_xrsimulator right?

okwasniewski: This is tricky because there shouldn't be x86 simulator (because x86 macs are not supported for visionOS but I think we might need also to build it for x86 and include it in the slice). 

When working with Hermes prebuilt, I encountered some issues when trying to upload the app to the App Store.  (https://github.com/facebook/react-native/pull/45911)

However at the moment there is no x86 build option in the `ios.toolchain.cmake`. Im going to add it and then contribute this back to https://github.com/leetal/ios-cmake

wcandillon: I am building it now, next should I just try to run the example app on the xrsimulator?

okwasniewski: Uhm, its not so simple unfortunately :D We would need to setup visionOS in the example app: https://callstack.github.io/react-native-visionos-docs/docs/getting-started/integrate-into-existing-projects 

Or create a new example. 

At this point visionOS doesn't support skia so we might need to also disable it for visionOS. Now that I got the build working Im going to try to integrate visionOS into the example and run it there

okwasniewski: I've integrated visionOS into the example app and it (almost) works, for some reason the `RectButton` was not working so I swapped it for TouchableOpacity and it works! 

Are you okay with keeping visionOS as an additional platform in the example app? (I can assist you with any updates / fixes regarding this platform)

wcandillon:
@okwasniewski Sorry that I didn't mention it earlier but today I migrated to a monorepo where the example app is an expo dev-client app. Could that work for you? Let me know.

wcandillon:
If you make a separate PR just for the build and so we have the prebuilt binaries + have them on github actions, I will merge it publish it asap.

okwasniewski:
Expo is unfortunately not supported (yet). If this is going to be a monorepo how about a separate project? I would love to run some visionOS CI to ensure that subsequential releases are not breaking the support as this library is crucial for that platform. Yeah, I can create a separate PR without the example

wcandillon:
yes that's a great idea :)

Also in the future all examples will be centralized in a single package
used by all apps. (depending on how fast I'm going with the current
migration).

wcandillon:
I think this is cool :) I just migrated to the monorepo. Could you add an visionpro in apps and with a simple example (e.g the red triangle). I will factorize the examples in its own package to be used in every `apps/` And potentially we can even run e2e tests on the vision pro simulator if supported.

Thank you so much for your hardwork. Hope the monorepo stuff is not throwing a wrench into your work.

okwasniewski:
> I think this is cool :) I just migrated to the monorepo. Could you add an visionpro in apps and with a simple example (e.g the red triangle). I will factorize the examples in its own package to be used in every `apps/` And potentially we can even run e2e tests on the vision pro simulator if supported.
> 
> Thank you so much for your hardwork. Hope the monorepo stuff is not throwing a wrench into your work.

No worries, I will add this! 

This might need to wait a little as I'm traveling for React Universe Conf next week but I will get back to this as soon as I can 👍🏻

okwasniewski:
Im going to close this one as the visionOS example added after https://github.com/wcandillon/react-native-webgpu/pull/109 gets merged (as react-native-test-app supports visionOS)

wcandillon:
that's exciting. So  react-native-test-app  is the way to go I didn't know about this thing

Saadnajmi:
@wcandillon We've switched a lot of React native libraries to use react-native-test-app as an easy way to keep maintenance and get out of tree platform support. visionOS and macOS for example get added with one file. My PR needs some work still, so I'd appreciate if anyone recognizes the errors I'm running into 😅

PR #89, opened 2024-08-26
fix: memory leak in metalViewRegistry

okwasniewski:
This PR fixes a memory leak that was coming from `metalViewRegistry`. 
 
Actually, it was never cleared 😅

In this PR I've changed:

- Assign `_contextId` of a given view in the update props method
- Check if key `_contextId` exists in `metalViewRegistry` and clear it properly.

Before this PR I've noticed some crashes and also metalViewRegistry was filled up with all the views that I've spawned.


wcandillon:
nice 😊😊😊

piaskowyk (approved):
🫡

PR #90, opened 2024-08-29
feat: visionOS build

okwasniewski:
This PR adds base for visionOS support. Example and CI will be added in future PRs


package/ios/SurfaceUtils.mm
```diff
@@ -10,7 +10,6 @@ + (void)configureSurface:(CALayer *)layer
                     size:(CGSize)size
                contextId:(int)contextId {
   std::shared_ptr<rnwgpu::RNWebGPUManager> manager = [WebGPUModule getManager];
-  CGFloat scaleFactor = [UIScreen mainScreen].scale;
```
okwasniewski, line 13: This was not used, and this API is not supported on visionOS so I just removed it. 

If there is a need for scaleFactor its preferable to retrieve it using [UITraitCollection](https://developer.apple.com/documentation/uikit/uitraitcollection?language=objc)

wcandillon: yes makes sense we don't do anything with this.

wcandillon:
very cool :)
Let me run the dawn build once on this branch and we should be good to go :)

wcandillon (approved):

wcandillon:
Now github complains about disk space 🥲 I guess I will split it into different machines 🤔

okwasniewski:
> Now github complains about disk space 🥲 I guess I will split it into different machines 🤔

You can also try using a large runner: https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/about-larger-runners#overview-of-larger-runners

The free limit for OSS should be enough

wcandillon:
@okwasniewski thanks a lot for the tip, here are the prebuilt: https://github.com/wcandillon/react-native-webgpu/actions/runs/10645260389 I will also publish them on npm 🫡

wcandillon:
@okwasniewski any chance you could try to contribute the changes to https://github.com/leetal/ios-cmake/blob/master/ios.toolchain.cmake?

PR #99, opened 2024-09-04
Fix relative paths for codegen after monorepo setup

kmagiera:
After monorepo setup has been added, the codegen stopped working because of relative paths refering to the locations from before the monorepo change.

This PR fixes the relative paths such that they refer to the new, updated location taking into account the changed structure of the repo.

### Test plan
1. run `yarn codegen` in `packages/webgpu` folder
2. before this change it'd complain about missing files
3. the script passes after applying this change


wcandillon:
nice thanks :)

wcandillon (approved):

PR #102, opened 2024-09-06
feat: Add macOS to build matrix

Saadnajmi:
Add macOS to the build matrix for dawn, and add the libraries to the `xcframework`. I can't get dawn building locally (most likely to my using Xcode 16 Beta). However, if this PR lands (and CI passes), then we'll have prebuilt macOS binaries available :).


wcandillon:
indeed the renaming from iOS to apple is needed now.

Saadnajmi:
Marking this as ready for review with a few caveats. I can't build dawn locally, most likely because I am using the Xcode 16 beta. I'll try with Xcode 15 later. Because of this, I'm not entirely sure the PR is accurate but comparing with #88 I think it looks mostly correct? And if this works on your machine (and Ci) then we also get prebuilt binaries from the CI artifacts 😀

packages/webgpu/react-native-wgpu.podspec
```diff
@@ -11,7 +11,7 @@ Pod::Spec.new do |s|
   s.license      = package["license"]
   s.authors      = package["author"]
 
-  s.platforms    = { :ios => min_ios_version_supported, :visionos => "1.0" }
+  s.platforms    = { :ios => min_ios_version_supported, :visionos => "1.0", :osx => "11.0" }
```
Saadnajmi, line 14: Actually, I should revert this. The rest of the sources don't compile for macOS yet.

packages/webgpu/scripts/build/dawn.ts
```diff
@@ -78,53 +80,67 @@ const ios = {
...
   libs.forEach((lib) => {
     console.log(`Building fat binary for visionos simulator: ${lib}`);
     $(
-      `lipo -create ${projectRoot}/libs/ios/x86_64_xrsimulator/${lib}.a ${projectRoot}/libs/ios/arm64_xrsimulator/${lib}.a -output ${projectRoot}/libs/ios/${lib}_visionos.a`,
+      `lipo -create ${projectRoot}/libs/apple/x86_64_xrsimulator/${lib}.a ${projectRoot}/libs/apple/arm64_xrsimulator/${lib}.a -output ${projectRoot}/libs/apple/${lib}_visionos.a`,
+    );
+  });
+
+  libs.forEach((lib) => {
+    console.log(`Building fat binary for macOS: ${lib}`);
```
wcandillon, line 114: This is not needed right?

packages/webgpu/scripts/build/dawn.ts
```diff
@@ -78,53 +80,67 @@ const ios = {
...
+        `-library ${projectRoot}/libs/apple/${lib}_visionos.a ` +
+        `-library ${projectRoot}/libs/apple/arm64_xros/${lib}.a ` +
+        ` -output ${projectRoot}/libs/apple/${lib}_visionos.xcframework `,
+    );
+     // macOS
+     $(
+      "xcodebuild -create-xcframework " +
+        `-library ${projectRoot}/libs/apple/${lib}_macos.a ` +
+        `-library ${projectRoot}/libs/apple/arm64_macosx/${lib}.a ` +
+        ` -output ${projectRoot}/libs/apple/${lib}_macos.xcframework `,
```
wcandillon, line 143: this probably should be (no fatlib for macos because there is no macos simulator):
```
    // macOS
    $(
      "xcodebuild -create-xcframework " +
        `-library ${projectRoot}/libs/apple/x86_64_macosx/${lib}.a ` +
        `-library ${projectRoot}/libs/apple/arm64_macosx/${lib}.a ` +
        ` -output ${projectRoot}/libs/apple/${lib}_macos.xcframework `,
    );
```

packages/webgpu/scripts/build/static_build.patch
```diff
@@ -8,7 +8,7 @@ index 046a6af10d..5a63ac3d6d 100644
  
 -  add_library(${output_target} SHARED ${all_objects})
 -
-+  if(${CMAKE_SYSTEM_NAME} STREQUAL "iOS" OR ${CMAKE_SYSTEM_NAME} STREQUAL "visionOS")
++  if(${CMAKE_SYSTEM_NAME} STREQUAL "iOS" OR ${CMAKE_SYSTEM_NAME} STREQUAL "visionOS" OR ${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")
```
wcandillon, line 11: 🙌🏼

wcandillon:
looking good. Due to the way podspecs work at the moment, I suggest that we build `UNIVERSAL` or `MAC_CATALYST_ARM64`. If we want to ship the arm64 to the x86 separately, we would need to taker a completely different approach.

wcandillon (approved):

PR #107, opened 2024-09-07
Three Fiber Example

wcandillon:


wcandillon:
@aarongrider this example runs out of the box. More complex examples don't run 😅 I think a good next step would be to first ensure that this approach works perfectly on Web first (which we are not sure at this point). And if it does, reconcilate the differences. Looking forward to read your thoughts on this.

aarongrider:
This is great, and much simpler than my PR!

Looks like the past few versions of three.js are causing issues with R3F (including on web). Currently investigating.

PR #110, opened 2024-09-09
Fabric custom shadow node for Android

piaskowyk:
This PR overrides the default shadow nodes generated by Fabric with custom ones to obtain information about component size before the native component is mounted. This will enable us to access information about the canvas size inside the `useEffect()`.

While we have successfully implemented this for iOS, for Android, I need to commit the files generated by the codegen to override the layout method since there is no available method `componentDescriptorProvider` like on iOS.

The StaticDataBridge library is utilized to transfer data between the Shadow Node and JNI module (cpp-adapter.cpp). This process is tricky because both files are compiled by different CMakeFiles.

This PR introduces lazy initialization of the View on Android. If the native surface isn't ready, it will produce an off-screen texture buffer, and once the surface is ready, the previously created texture will be pushed to the native surface.


wcandillon:
@piaskowyk I am closing this in favor of #137 but let's keep refine this and build the best Canvas API in React Native 🤝

PR #112, opened 2024-09-10
TypeGPU gradient tiles example

mhawryluk:
Basic example showing the TypeGPU library usage for WebGPU rendering.


wcandillon (changes requested):
I love the example. However, I am not seeing TypeGPU has a separate category of example. We will have many different examples in the future and most likely they will use TypeGPU more and more in the future. Therefore I recommend you simply add the GradientTiles example in the regular list.

This will put an emphasis on the fact that TypeGPU is not a separate thing but just a very helpful way to write WebGPU code.

apps/paper/ios/Podfile.lock
```diff
@@ -1573,7 +1573,7 @@ SPEC CHECKSUMS:
   RNReanimated: abb16e70ea4fbd72c667946caa3abd0f962897f6
   RNScreens: aa943ad421c3ced3ef5a47ede02b0cbfc43a012e
   SocketRocket: abac6f5de4d4d62d24e11868d7a2f427e0ef940d
-  Yoga: ae3c32c514802d30f687a04a6a35b348506d411f
+  Yoga: 2f71ecf38d934aecb366e686278102a51679c308
```
wcandillon, line 1576: Where does this change come from?

wcandillon (approved):

apps/paper/ios/Podfile.lock
mhawryluk: it changed on its own when I ran `pod install`

PR #118, opened 2024-09-12
TypeGPU Compute Boids example

reczkok:
Boids simulation based on [the Compute Boids example](https://webgpu.github.io/webgpu-samples/?sample=computeBoids) with some behavior and color presets.

https://github.com/user-attachments/assets/328691e9-d5cb-4634-aea6-63842c0ebf33


wcandillon:
this is very cool 💚

PR #109, opened 2024-09-09
Replace test app with `react-native-test-app`

Saadnajmi:
DRAFT, DO NOT MERGE (Figured I'd throw up my WIP branch anyway)

In order to  help maintenance and improve support for new out of tree platforms (like macOS and visionOS), let's replace the test app with https://github.com/microsoft/react-native-test-app . This simplifies the native code needed to stand up the example apps. In the case of apple platforms, it's now a single pod-sec.

TODO:
- I had to port some of the manual resolving of the node module `three/webgu`, particularly in `metro.config.js`. We should double check that before merge.

- I am still getting the following error with the latest set of my changes when I try to load from dev server:
```
 BUNDLE  ./index.js 

 LOG  Running "Example" with {"rootTag":61,"initialProps":{}}
 ERROR  Invariant Violation: TurboModuleRegistry.getEnforcing(...): 'WebGPUModule' could not be found. Verify that a module by this name is registered in the native binary.Bridgeless mode: false. TurboModule interop: false. Modules loaded: {"NativeModules":["PlatformConstants","LogBox","Timing","AppState","SourceCode","BlobModule","WebSocketModule","SettingsManager","DevSettings","RedBox","Networking","Appearance","DevLoadingView","UIManager","DeviceInfo","ImageLoader","LinkingManager"],"TurboModules":[],"NotFound":["NativePerformanceCxx","NativePerformanceObserverCxx","BugReporting","HeadlessJsTaskSupport","SoundManager","IntentAndroid","WebGPUModule"]} 
loadModuleImplementation@http://localhost:8081/index.bundle//&platform=ios&dev=true&lazy=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&app=com.microsoft.ReactTestApp:267:14
RCTView
View@http://localhost:8081/index.bundle//&platform=ios&dev=true&lazy=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&app=com.microsoft.ReactTestApp:63675:43
RCTView
View@http://localhost:8081/index.bundle//&platform=ios&dev=true&lazy=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&app=com.microsoft.ReactTestApp:63675:43
AppContainer@http://localhost:8081/index.bundle//&platform=ios&dev=true&lazy=true&minify=false&inlineSourceMap=false&modulesOnly=false&runModule=true&app=com.microsoft.ReactTestApp:63548:25
[...]

PR #120, opened 2024-09-17
Replace test app with react-native-test-app

okwasniewski:
In order to help maintenance and improve support for new out of tree platforms (like macOS and visionOS), let's replace the test app with https://github.com/microsoft/react-native-test-app . This simplifies the native code needed to stand up the example apps. In the case of apple platforms, it's now a single pod-sec.

Most of the work was done here: #109 by @Saadnajmi so kudos to him for that! 

CleanShot 2024-09-17 at 17 11 12@2x


okwasniewski:
TODO: Fix the CI 😅

Going to do this tomorrow

wcandillon:
nice 🫡 let me know if I can help with the CI

wcandillon:
@okwasniewski  silly question, does it support RN Web as well? Sorry I know nothing about this topic yet but it sounds very cool

okwasniewski:
@wcandillon So unfortunately, there is no first class support for web in test-app: https://github.com/microsoft/react-native-test-app/issues/812

I guess you could have two examples (test-app + expo) that would probably give you the best coverage

okwasniewski:
I found an issue with the integration of test-app + turbo

I've contacted @tido64 so we can come up with a solution. 

TLDR of the issue: 

When running `yarn react-native build-ios` test app generates a new `.xcode.env` file and executes: `which node` command to supply it for Xcode. 

However when executing this command through turbo (https://turbo.build/)

`which node` returns: `[DEBUG] output from which node: /private/tmp/xfs-1a72c2db/node` <- not correct path

Making the Xcode build fail

okwasniewski:
Hey @wcandillon build is now green!

Its ready to review

wcandillon:
haha no way :)

wcandillon:
silly question: can we easily build the app in the new arch? 😇

okwasniewski:
> silly question: can we easily build the app in the new arch? 😇

Yeah! For sure. 

For iOS: 

`RCT_NEW_ARCH_ENABLED=1 pod install --project-directory=ios`

For Android: 

Modify `apps/example/android/gradle.properties`:

```
newArchEnabled=true
```

We can also add new arch build to the CI later

wcandillon:
this is fantastic thank. Right now this doesn't work on iOS (Canvas API crashes, that's a bug on our side) nor Android (e2e not working, probably an issue on our side as well). However are currently fixing a serious bug in our Canvas API.

What I will do on my side is to work on shipping this bug fixes asap so we can test it. But I promise I won't touch the example app so you don't need to fix time fixing conflicts and so on.

Thank you again for this, this is very compelling.

Saadnajmi:
Awesome work, thanks @okwasniewski for taking the change and getting it ready and @tido64 for helping RNTA work with Turborepo repos!

wcandillon:
None of the examples work for me. First I thought it was related to some canvas API bugs we are currently fixing but now it looks like it might be something else?

okwasniewski:
@wcandillon What is the error you are seeing? Im taking a look

okwasniewski:
@wcandillon Im sorry but I can't reproduce this issue, everything works fine for me. Just to be sure you disabled the Metal API Validation in Xcode? 


https://github.com/user-attachments/assets/8521253d-7cbd-411b-a913-d4382a48059c

wcandillon (approved):
this is big thank you for this.

okwasniewski:
@wcandillon Can you hint me what is crashing for you? I can see the build is failing but that's because of missing dawn.json file.

wcandillon:
the e2e tests but the same error happens when browsing the examples. I also cannot install the apk on Android but the issue might be on my side there.

okwasniewski:
@wcandillon I'm going to test Android soon, so the tests were green on previous app and now they are red? Is it the setup of the tests or is it a native crash? I didn't run them

wcandillon:
It's a native crash that I can also reproduce by browsing the examples.

Saadnajmi:
I'll also try and test

okwasniewski:
@wcandillon Sorry for the delay. 

I went to every example and no crash (I built using Android Studio) but it shouldn't matter. Can you provide a crash stack trace? Either way I don't think its related to TestApp more likely its the library.

https://github.com/user-attachments/assets/06d34204-45d3-4d26-85cf-b48613353249

wcandillon:
Here is the list of issues/regressions I am experiencing. However, these probably shouldn't be called regressions since we are just changing the example app, so the module should still work there.

On Android, everything appears to work properly except for the `<Texture />` used in tests. There might be a bug on my side here, so I think we can let this one go and fix it later.

On iOS, the `ExternalTexture.spec.ts` e2e test fails, and running the three.js helmet example followed by the r3f example causes a crash.

This behavior is not happening in apps/paper, and I'm not sure why yet.

Saadnajmi:
> Here is the list of issues/regressions I am experiencing. However, these probably shouldn't be called regressions since we are just changing the example app, so the module should still work there.
> 
> On iOS, the `ExternalTexture.spec.ts` e2e test fails, and running the three.js helmet example followed by the r3f example causes a crash.
> 
> This behavior is not happening in apps/paper, and I'm not sure why yet.

@wcandillon is there any chance those issues are intermittent and not related to this PR? I can't think of why the test app would cause those things to fail, and I would personally like this PR to unblock new architecture / macOS / potential future Windows support. I just set up a new laptop so I'll try to repro the issues mentioned sometime this week on a clean install.

wcandillon:
I found the issue and it is as we expected unrelated to this PR, sorting it out now.

Saadnajmi:
> @okwasniewski do you know how I can have the "Metal API Validation" unchecked in the example app? If I uncheck it in xcode, I don't see a change in git, probably because the file is in gitignore?

@tido64 perhaps this is a feature request for RNTA?

wcandillon:
@okwasniewski I am also unable to build the app on fabric/android, my bad for not testing it before. 
@Saadnajmi Any specific instructions when looking to upgrade the example app?

okwasniewski:
@wcandillon There you go: https://github.com/microsoft/react-native-test-app/wiki/Platform-Specifics:-Android#new-architecture-previously-fabric-turbomodule-or-bridgeless

tido64:
> @okwasniewski do you know how I can have the "Metal API Validation" unchecked in the example app? If I uncheck it in xcode, I don't see a change in git, probably because the file is in gitignore?

You will have to write an Expo config plugin for this (or patch the `.xcscheme` some other way). You should find the `.xcheme` at `node_modules/.generated/ios/ReactTestApp.xcodeproj/xcshareddata/xcschemes/<MyApp>.xcscheme`, and you need to set `enableGPUValidationMode = "0"` on the first `LaunchAction`.

More info on Config Plugins can be found here: https://github.com/microsoft/react-native-test-app/wiki/Config-Plugins

PR #123, opened 2024-09-19
feat: Add macOS support

Saadnajmi:
Add macOS support to `react-native-webgpu`. This is done by:

- Moving the apple sources from `ios/*` to `apple/*`
- Adding macOS to the podspec
- Adding a new `RNWPUIKit.h` that adds a typedef for `NSView` and `UIView` so we have less ifdefs in the codebase ([Reanimated](https://github.com/software-mansion/react-native-reanimated/blob/main/packages/react-native-reanimated/apple/reanimated/REAUIKit.h) and [GestureHandler](https://github.com/software-mansion/react-native-gesture-handler/blob/b79223de5a47773a464543018b7231113c1ade68/apple/RNGHUIKit.h) do the same thing)
- Rename `IOSPlatformContext` to `ApplePlatformContext`
- Add some ifdefs for macOS specific code
- Switch from `@react-navigation/native-stack` to `@react-navigation/stack` as the former does not have macOS support.

Screenshot 2024-09-23 at 11 26 35 AM


Saadnajmi:
@wcandillon Does this package require Hermes? I was getting issues running with JSC. Once #120 lands, I can rebase this PR and see if I'm still seeing those errors :)

Saadnajmi:
Got it working on macOS with an additional ifdef to get `CAMetalLayer` working! I may add the needed directly to RNM instead of having the ifdef here :)

Saadnajmi:
@wcandillon Could you let me know if this runs on your machine? I just rebased off of main

apps/example/package.json
```diff
@@ -16,16 +16,14 @@
...
-    "@react-three/fiber": "^8.17.6",
+    "@react-navigation/stack": "^6.4.0",
     "fast-text-encoding": "^1.0.6",
     "react": "18.2.0",
     "react-native": "0.74.2",
     "react-native-gesture-handler": "^2.17.1",
     "react-native-macos": "^0.74.0",
     "react-native-reanimated": "^3.12.1",
     "react-native-safe-area-context": "^4.10.7",
-    "react-native-screens": "^3.32.0",
```
Saadnajmi, line 28: and this one..

Saadnajmi:
I see something with @react-three/fiber is failing, I'm trying https://github.com/wcandillon/react-native-webgpu/pull/140 to see if it helps.

wcandillon:
@Saadnajmi I like it a lot thanks for doing it. Anything missing? would you like to add a build step for it on the ci.yml?

wcandillon:
@Saadnajmi I'm not able to compile the project (reanimated and gesture-handler are not compiling, am I missing something?
Here's one of the error I'm getting:
Screenshot 2024-10-15 at 21 37 16

Saadnajmi:
> @Saadnajmi I'm not able to compile the project (reanimated and gesture-handler are not compiling, am I missing something? Here's one of the error I'm getting: Screenshot 2024-10-15 at 21 37 16

OK, I thought I fixed that in React Native macOS, perhaps we need to update to the latest patch version. Let me try locally.
I can add a step for it in CI. I wasn't sure if we wanted to commit to that yet :)

wcandillon:
looks like you need to update the turbo.json and the package.json at the root for build:macos to work

Saadnajmi:
> looks like you need to update the turbo.json and the package.json at the root for build:macos to work

I think "react-native build-macos" isn't part of the CLI. I can't remember if we have "react-native-macos build-macOS" either. 

There's some outstanding CLI work we need to do to make sure our commands are registered (ever since the React Native Community CLI got decoupled from React Native) so this PR might get blocked on that :/

wcandillon (changes requested):
thanks for the update. it looks like the MetalView needs to be updated with the following:
```
#if !TARGET_OS_OSX
- (void)willMoveToSuperview:(UIView *)newSuperview {
  [super willMoveToSuperview:newSuperview];

  if (newSuperview == nil) {
    // The view is being removed from its superview
    // Add your cleanup code here
    [SurfaceUtils cleanupSurface:[_contextId intValue]];
    _isConfigured = NO;
  }
}
#else // macOS specific lifecycle method
- (void)viewWillMoveToSuperview:(NSView *)newSuperview {
  [super viewWillMoveToSuperview:newSuperview];

  if (newSuperview == nil) {
    // The view is being removed from its superview
    // Add your cleanup code here
    [SurfaceUtils cleanupSurface:[_contextId intValue]];
    _isConfigured = NO;
  }
}
#endif
```

Gesture handler is not working for me but I guess this is expected? If yes that's more than fine. We could work around it.

The examples where not working for me but also I am not sure how you were able to build the app or run the examples? I would be interested to learn more here.

Could you look at `globalSetup.ts` and update the platform checks: https://github.com/wcandillon/react-native-webgpu/blob/main/packages/webgpu/src/__tests__/globalSetup.ts#L14

apps/example/package.json
```diff
@@ -8,15 +8,17 @@
     "android": "react-native run-android",
     "ios": "react-native run-ios",
     "start": "react-native start",
-    "pod:install": "pod install --project-directory=ios",
+    "pod:install:ios": "pod install --project-directory=ios",
+    "pod:install:macos": "pod install --project-directory=macos",
     "build:android": "cd android && ./gradlew assembleDebug --warning-mode all",
     "build:ios": "yarn run mkdist && react-native bundle --entry-file index.js --platform ios --dev true --bundle-output dist/main.ios.jsbundle --assets-dest dist && react-native build-ios --scheme Example --mode Debug --extra-params \"-sdk iphonesimulator CC=clang CPLUSPLUS=clang++ LD=clang LDPLUSPLUS=clang++ GCC_OPTIMIZATION_LEVEL=0 GCC_PRECOMPILE_PREFIX_HEADER=YES ASSETCATALOG_COMPILER_OPTIMIZATION=time DEBUG_INFORMATION_FORMAT=dwarf COMPILER_INDEX_STORE_ENABLE=NO\"",
+    "build:macos": "yarn run mkdist && react-native bundle --entry-file index.js --platform macos --dev true --bundle-output dist/main.macos.jsbundle --assets-dest dist",
```
wcandillon, line 15: kudos :))

wcandillon: that doesn't seem to build the app, could we put the xcode build command here?

Saadnajmi: Yeah, I think I need to actually go and make sure that shows up in the React Native community CLI :D. I'll follow up on that, and meanwhile split out some more changes from this PR.

Saadnajmi: I checked with @tido64. We need to go and implement `build-macos`. We have `run-macos`... so it shouldn't be too hard. But it might be a bit.

Saadnajmi:
I've updated the PR after my other commits landed, so the macOS native changes are quite minimal now. Running locally, most of the examples fail with " ERROR  Error: Exception in HostFunction: Not implemented", not sure what changed there :(.


Screenshot 2024-10-17 at 12 57 33 PM

wcandillon:
@Saadnajmi Thank you for this. I will try it. I see that the CI is failing now?

Saadnajmi:
> @Saadnajmi Thank you for this. I will try it. I see that the CI is failing now?

Yeah, it'll fail till we go implement `build-macos`. If that's not a blocker, I'll remove the CI changes and separate. But it's on our backlog now so we're looking into it.

Saadnajmi:
@wcandillon looks like CI works now!

wcandillon (approved):
Looks good we can merge/release it. Thank you for your efforts on this 🤝

wcandillon:
@Saadnajmi I broke the build, fixing it now.

PR #124, opened 2024-09-20
fix: properly build xcframework

okwasniewski:
Hey, 

This PR features multiple changes to fix how `xcframeworks` are handled. 

Changes included: 
- Move from `macos-latest-large` to `macos-latest` for the CI (Only `build-dawn.yml` requires M1 machine for now)
- Remove visionOS x86 build (building on Intel macs for visionOS is deprecated so lets not waste resources on this)
- I aligned `apple.toolchain.cmake` with upstream source code.
- Package only `xcframeworks` in build dawn job (currently the artifacts were over 450mbs because contained unnecessary stuff)

### Changes to build scripts

So before we created a bunch of xcframeworks then linked them for each platform. 

The whole idea behind xcframeworks is to have one bundle for multiple platforms(!). 

Before (contents generated by build-dawn job): 

CleanShot 2024-09-20 at 14 36 12@2x

Now: 

CleanShot 2024-09-20 at 14 35 31@2x

PR #125, opened 2024-09-23
fix: ci build

okwasniewski:
This PR fixes the CI build to check if the build fail is caused by #124 or if is it failing in general

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,7 @@ jobs:
         run: yarn build
 
   build-android:
-    runs-on: macos-latest-large
+    runs-on: macos-latest
     env:
       TURBO_CACHE_DIR: .turbo/android
     steps:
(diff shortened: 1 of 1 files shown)


wcandillon:
@okwasniewski I think it's because I ran over my github credits 🥲

okwasniewski:
> @okwasniewski I think it's because I ran over my github credits 🥲

@wcandillon Yeah, that's the case 😅 But we don't need these more expensive workers for this build so its fine to use normal ones. I changed the workers on my other PR and it was still failing. I want to check if this is failing because of my PR or there is a general build issue on the CI

okwasniewski:
@wcandillon the build is green I guess we can merge it and I will figure out what breaks it in my other PR. WDYT?

wcandillon:
yes 🫡 Sorry we've been slow with the other PRs it's just that we are working on fixing substancial bugs in canvas API and I I would be more at ease if these would be merged to test the other PRs

wcandillon (approved):

PR #129, opened 2024-09-24
Fix: `await device.lost` hangs when called before `device.destroy()`

### LINKED ISSUE #126: `await GPUDevice.lost` hangs indefinitely if called before `GPUDevice.destroy()`
Hello!
I'm super interested in adopting `react-native-webgpu` for an app I'm working on. Since the library is quite new, I've been trying to run the [gpuweb compliance test suite](https://github.com/rebeckerspecialties/gpuweb-cts) against the library to gather an inventory of implementation gaps and gotchas, but there have been a couple of blockers. 

The major blocker for running the compliance test suite is that some unresolved promises hang the main thread and cannot be cancelled. For example, this code (split out from [this test](https://github.com/gpuweb/cts/blob/815ff2bb4038144dea89c33021bc4429f22a130f/src/webgpu/api/operation/adapter/requestDevice.spec.ts#L57-L87)) should time out after 1 second, but it hangs the app instead (the panresponder becomes unresponsive):
```ts
const adapter = await navigator.gpu.requestAdapter();
const device = await adapter.requestDevice();

const timeoutPromise = new Promise((_resolve, reject) => {
    const handle = setTimeout(() => {
      reject(new Error('timeout'));
    }, 1000);
});

// device.lost never resolves or rejects, hanging the app
await Promise.race([device.lost, timeoutPromise]);
```

It seems like the issue occurs when converting from `std::future` to a JS `Promise` in the [RNFJSCIConverter](https://github.com/wcandillon/react-native-webgpu/blob/main/packages/webgpu/cpp/jsi/RNFJSIConverter.h#L210-L243). The code blocks the main thread on `sharedFuture->wait();`, which never resolves for `device.lost`, since `device.destroy` has not been called. The original author of the JSI converter code has [since updated it](https://github.com/mrousavy/nitro/blob/main/packages/react-native-nitro-modules/cpp/jsi/JSIConverter%2BPromise.hpp#L42-L86) to add some additional async handling. Would the ThreadPool approach used there make sense for this library? 

I would be happy to submit a PR if that approach seems appropriate to you. There are a handful of other cases in the compliance test suite that this would resolve. If you're interested in getting the inventory from the compliance test suite, I'd be happy to file more issues as I find them and submit fixes! The other issues I've noticed result from the Dawn version trailing the latest Chromium version, but I can file a separate issue.

Thanks for the consideration and creating this awesome library!

hmallen99:
Resolves #126 

## Problem
Fixes a bug where javascript promises will hang the main thread while waiting for an `std::future` value on the native side. In the case of `await device.lost`, this causes the main thread to hang indefinitely, as `device.destroy` cannot be called to resolve the promise. This bug was discovered by running the impl against the [gpuweb cts](https://github.com/rebeckerspecialties/gpuweb-cts).

## Solution
- Updates the conversion from `std::future` to JS promise to use a more robust async implementation, allowing promises to await on a background thread.
- Copies the threading code from [react-native-nitro-modules](https://github.com/mrousavy/nitro/tree/main/packages/react-native-nitro-modules/cpp/threading), with some logging and debug code stripped out. `.hpp` files have been converted to `.h` files and the namespace has been shortened from `margelo::nitro` -> `margelo` to match the conventions used in this repo.

## Testing
- Added a unit test in the `Device` test suite to validate that device.lost will not hang the main thread if device.destroy is not called. With the old threading logic, this test hangs indefinitely. With the new test logic, we are able to cancel the promise.
- Added a unit test to validate that `device.lost` resolves if it is awaited before calling `device.destroy`
- After doing some more testing against the gpuweb-cts, all of the promise hangs resulted from one of these two scenarios


wcandillon:
@hmallen99 I'm very excited about this. 

Right now I'm heads down with some deep Canvas API refactorings but after that I could look into the nitro module migration.

hmallen99:
> @hmallen99 I'm very excited about this.
> 
> Right now I'm heads down with some deep Canvas API refactorings but after that I could look into the nitro module migration.

Migrating to nitro modules would be ideal! I think this should act as a stop-gap until that migration can happen, though. 

I've tested this against the conformance test suite, and it removes a lot of the blockers for running the tests. The remaining inventory seems to be around throwing the proper `RangeError` for incorrect buffer sizes and catching some shader compilation issues. I can file some more issues!

wcandillon:
I agree! Should we jump on a call to discuss how to run CTS with this module? I would love to learn more.

hmallen99:
Yes, let's hop on a call @wcandillon! Please let me know how and when you would like to call. The process of running tests requires a bit of setup right now, but I can walk you through some of it. It's fairly similar to running the example app in this repo. I'm hoping to make it more automated in the future, though.

hmallen99:
I also updated the README to include build and test instructions in my fork of the cts, if you want to give that a try. It would be useful to have the feedback if those instructions don’t make sense! https://github.com/rebeckerspecialties/gpuweb-cts/blob/main/README.md

wcandillon:
my email is public, can you send me a proposal calendar invite?

hmallen99:
> my email is public, can you send me a proposal calendar invite?

Sent you an invite!

hmallen99:
Hey @wcandillon, are we able to merge this PR? It should be working on both iOS and android with tests passing. Thanks!

wcandillon:
thanks for letting me know, I will start reviewing it.

hmallen99:
> thanks for letting me know, I will start reviewing it.

Hmm, I'll take a look at the CI failures. Thanks for running the actions!

wcandillon:
@hmallen99 we're on the verge of merging #137, this PR will be next immediately after that 🙏

hmallen99:
> @hmallen99 we're on the verge of merging #137, this PR will be next immediately after that 🙏

Awesome, I just validated that all the static analysis and test steps are passing locally. I'll validate the `build-android` and `build-ios` steps as well.

wcandillon:
@hmallen99 just had a look and it makes sense . Did you try it on Android? I don't think we pass the jsInvoker there.

wcandillon:
@hmallen99 I can fix the android issue for you

hmallen99:
> @hmallen99 I can fix the android issue for you

Thanks @wcandillon, that would be great! The android simulator seems to run very slowly on my machine, so it's hard for me to debug.

wcandillon:
@hmallen99 I've merge the latest main including having the callinvokator on android. Could you do some testing our sides? I'm having issues on both iOS and Android now 🤔

hmallen99:
> Could you do some testing our sides? I'm having issues on both iOS and Android now 🤔

It should be fixed on iOS now, I'll take a look at Android as well. It looks like the thread pool isn't being registered properly, even with the call invoker added. 

Also, I tried to uptake the latest commit with the dawn update, but it seems to be causing some build issues, so I've reverted it in this branch for now!

hmallen99:
The issue may be localized to the simulator -- I was able to get the example app running on an Android device. Does that match what you're seeing? @wcandillon

wcandillon:
I've upgraded Dawn you need to rebuild it locally or download the new build artifacts: https://github.com/wcandillon/react-native-webgpu?tab=readme-ov-file#downloading-dawn

wcandillon (approved):
This works well, thank you for your hard word on this :) Hope we can even go deeper with the CTS

hmallen99:
> This works well, thank you for your hard word on this :) Hope we can even go deeper with the CTS

Thanks! I'll try to run the CTS again this week and report back with some results :). Excited to do more with this library.

mrousavy:
Instead of copying in more code from Nitro, a full switch to Nitro Modules directly would be a much better option.

Even if it's more work, it's more robust, and shares code with other Nitro Modules.

Also it'd benefit from all the countless improvements I've made to Nitro so far - RN WebGPU still has the old HybridObject (`jsi::HostObject`) implementation which is magnitudes slower than the one I have in Nitro `main`.

wcandillon:
very good point indeed and this part will be removed quite soon. It doesn’t
work on worklet thread (it always try to callback the promise on the js
thread).  I‘ll definitely check out nitro modules see if it fits, this part
and in Skia will be revamped soon. So far we‘ve been focused on other
performance/stability aspects of these modules.

PR #137, opened 2024-10-07
Improvements to the Canvas API

wcandillon:


packages/webgpu/android/cpp/cpp-adapter.cpp
```diff
@@ -27,52 +27,55 @@ extern "C" JNIEXPORT void JNICALL Java_com_webgpu_WebGPUModule_initializeNative(
...
-    JNIEnv *env, jobject thiz, jobject surface, jint contextId, jfloat width,
+    JNIEnv *env, jobject thiz, jobject jSurface, jint contextId, jfloat width,
     jfloat height) {
-  auto window = ANativeWindow_fromSurface(env, surface);
+  auto window = ANativeWindow_fromSurface(env, jSurface);
   // ANativeWindow_acquire(window);
-  manager->surfacesRegistry.addSurface(contextId, window, width, height);
+  auto &registry = rnwgpu::SurfaceRegistry::getInstance();
+  // 1. The scene has already be drawn offscreen
+  if (registry.hasSurface(contextId)) {
```
wcandillon, line 53: hasSurfaceInfo

packages/webgpu/cpp/rnwgpu/api/GPUCanvasContext.cpp
```diff
@@ -23,41 +23,98 @@ void GPUCanvasContext::configure(
...
-  _instance.Configure(&surfaceConfiguration);
-  _lastConfig = configuration;
-  _width = _canvas->getWidth();
-  _height = _canvas->getHeight();
+  auto width = _canvas->getWidth();
+  auto height = _canvas->getHeight();
+  surfaceConfiguration.width = width;
+  surfaceConfiguration.height = height;
+  _surfaceConfiguration = surfaceConfiguration;
+  _offscreenSurface->configure(_surfaceConfiguration);
```
wcandillon, line 31: Check if we need to configure the offscreen (e.g. if when calling configure, the onscreen surface is already available)

packages/webgpu/cpp/rnwgpu/api/GPUCanvasContext.cpp
```diff
@@ -23,41 +23,98 @@ void GPUCanvasContext::configure(
...
+      auto info = registry.getSurface(_contextId);
+      // if the native surface is available, but we didn't create the WGPU instance yet, do it now
+      if (info.surface == nullptr) {
+        info.surface = _platformContext->makeSurface(_gpu->get(), info.nativeSurface, width, height);
+      }
+      _instance = info.surface;
+
+      _surfaceConfiguration.width = width;
+      _surfaceConfiguration.height = height;
+      _instance.Configure(&_surfaceConfiguration);
```
wcandillon, line 73: check with comment above

packages/webgpu/cpp/rnwgpu/api/GPUCanvasContext.cpp
```diff
@@ -23,41 +23,101 @@ void GPUCanvasContext::configure(
...
+  }
 }
 
 void GPUCanvasContext::unconfigure() {
-  _lastConfig = nullptr;
-  _width = _canvas->getWidth();
-  _height = _canvas->getHeight();
-  _instance.Unconfigure();
+  if (_instance) {
+    _instance.Unconfigure();
```
piaskowyk, line 51: I'm just wondering, what if `_offscreenSurface` was already configured? Should you also unconfigure it here, or is it not necessary?

packages/webgpu/cpp/rnwgpu/SurfaceRegistry.h
```diff
@@ -1,33 +1,159 @@
...
+  void *nativeSurface = nullptr;
+  wgpu::Surface surface;
+  int width;
+  int height;
+  wgpu::Texture texture;
+  wgpu::Instance gpu;
+  wgpu::SurfaceConfiguration config;
+
+  void flush() {
+    // 1.a flush texture to the onscreen surface
```
piaskowyk, line 26: This might be a bit confusing, why does the `SurfaceInfo` struct contain a method `flush` that works only for offscreen texture

piaskowyk (commented):
Good job! 🎉 

Just one small thing - I think it would be helpful if you could add a JS exception when someone tries to use the context before knowing the size of the texture. Showing a RedBox error would be more informative for users than just experiencing a native crash, in my opinion 🤔

packages/webgpu/cpp/rnwgpu/api/GPUCanvasContext.cpp
wcandillon: it's not necessary but I will clarify this indeed.

packages/webgpu/src/Canvas.tsx
```diff
@@ -31,45 +44,92 @@ export interface NativeCanvas {
...
+        cb.current = callback;
+      },
+      getCallback: () => {
+        return cb.current;
+      },
+      getContext(contextName: "webgpu"): RNCanvasContext | null {
+        if (contextName !== "webgpu") {
+          throw new Error(`[WebGPU] Unsupported context: ${contextName}`);
+        }
+        if (size === null) {
```
wcandillon, line 119: here it throws an error if the size is not available yet. @piaskowyk is that what you meant? I should add it to getNativeSurface too

PR #140, opened 2024-10-09
chore: Use Yarn 4 to patch packages instead of patch-package

Saadnajmi:
Yarn 2+ supports patching packages directly on `yarn install`, and has the added benefit that it fixes the version in your package.json so you don't accidentally bump the package.


wcandillon (approved):

wcandillon:
Thank you for this :)

PR #142, opened 2024-10-13
Replace SurfaceView with TextureView

Yaski:
It works but is not optimal because I have to create a new Surface every time SurfaceTexture changes. I have to convert to Surface because the native methods in webgpu only accept input as Surface.
Because of this there is flickering in "Resize" example.

### DIFF
--- a/packages/webgpu/android/src/main/java/com/webgpu/WebGPUView.java
+++ b/packages/webgpu/android/src/main/java/com/webgpu/WebGPUView.java
@@ -3,21 +3,21 @@ package com.webgpu;
 import androidx.annotation.NonNull;
 
 import android.content.Context;
+import android.graphics.SurfaceTexture;
 import android.view.Surface;
-import android.view.SurfaceHolder;
-import android.view.SurfaceView;
+import android.view.TextureView;
 
 import com.facebook.proguard.annotations.DoNotStrip;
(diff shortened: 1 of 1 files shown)


wcandillon:
@Yaski Thank you for this PR, we're considering other approaches for transparency on Android, stay tuned.

wcandillon:
It doesn't solve the issue right? I'm currently investigating this topic.

Yaski:
@wcandillon using TextureView solves issue with black background. Are you mentioning some other issue?
And I tried to make background transparent and I couldn't for some reason.

wcandillon:
@Yaski I've done more testing and we are shipping transparency on iOS now. For Android, the current plan will be to keep surface view as the default, offer TextureView via a prop (named `<Canvas androidTransparency />` or something) and for Android 29 and above, use an Android hardware buffer. Please bare with us as we work on these issues.

wcandillon:
@Yaski Thank you for your implementation. We are offering transparency via the `androidTransparency`. Right now we offer SurfaceView (opaque), TextureView (animates slower), and AHB (API Level => 29, animates faster but resizing is slightly slower at the moment. I am thinking of releasing first just the SurfaceView and Texture implementation and enable the AHB implementation later.

I am closing this in favor of #168

wcandillon:
We shipped this now :)

PR #143, opened 2024-10-13
Fix surface lifecycle on Android

### LINKED ISSUE #135: Black canvas (and crash) after going from background to foreground on Android
After using webgpu canvas with skia i've noticed that when you switch to another app and then go back to your app where you have webgpu canvas showing - it will become blank.

I've confirmed this same problem in the example app, here is recording.

Pixel 8 pro running Android 14
webgpu 0.1.8

in our app we use RN 0.73

https://github.com/user-attachments/assets/fe0d9f85-5356-42b6-a043-d6077bda5723

wcandillon:
This fixes #135 but is also a huge cleanup/refactor of #137


apps/example/src/Home.tsx
```diff
@@ -27,6 +27,10 @@ export const examples = [
     screen: "Cube",
     title: "🧊 Cube",
   },
+  // {
```
matthargett, line 30: Did you mean to leave this commented out?

wcandillon: yes I just use this for debugging. Maybe we could enable it and show how the CanvasAPI can be used to run Skia on top of it.

PR #154, opened 2024-10-17
⬆️ Upgrade to 0.75

wcandillon:


wcandillon:
@Saadnajmi @okwasniewski Any chance you could help with the upgrade? I read the documentation and I am not sure what I should do? Compile error on the WebGPU side I can fix but it looks like another module is failing?

Saadnajmi:
> @Saadnajmi @okwasniewski Any chance you could help with the upgrade? I read the documentation and I am not sure what I should do? Compile error on the WebGPU side I can fix but it looks like another module is failing?

I hadn't seen that particular module fail, I can try locally and see what's up next week perhaps?

wcandillon:
I'm getting the same error locally

Saadnajmi:
Hey, just as an FYI, React Native 0.76 is out for Android, iOS, and macOS. I can look at this PR later and see what's up, but I'm curious if you'd rather update to 0.76?

wcandillon:
I would definitely upgrade to 0.76 preferably ☺️

PR #158, opened 2024-10-20
Flush onscreen texture to offscreen texture on context switches

wcandillon:


wcandillon:
not valid since is the content of the current texture has been presented already

PR #162, opened 2024-10-22
Fix `undefined symbol: facebook::react::CallInvokerHolder::getCallInvoker()`

piaskowyk:
Fix for: `ld: error: undefined symbol: facebook::react::CallInvokerHolder::getCallInvoker()`

Now applications builds without errors:
image


wcandillon:
thank you so much :) :)

PR #187, opened 2024-12-10
Fix Android build and linking on RN 0.76

mhawryluk:
When trying to run an app using the library on Android with React Native 0.76 I encountered two issues:

1.

```
WebGPUModule.java:22: error: WebGPUModule is not abstract and does not override abstract method createSurfaceContext(double) in NativeWebGPUModuleSpec
public class WebGPUModule extends NativeWebGPUModuleSpec {
```

The `createSurfaceContext` method implementation was indeed removed in #137 

2.
```
  CMake Error at CMakeLists.txt:26 (add_library):
    Target "react-native-wgpu" links to target "ReactAndroid::reactnativejni"
    but the target was not found.
```

This is likely because of the native library merge in 0.76: https://reactnative.dev/blog/2024/10/23/release-0.76-new-architecture#android-apps-are-38mb-smaller-thanks-to-native-library-merging

I think this should fix both of the errors. It works for me on React Native 0.76. However, I didn't manage to test it on older versions, as I get seemingly unrelated errors even without any changes.

### DIFF
--- a/packages/webgpu/android/CMakeLists.txt
+++ b/packages/webgpu/android/CMakeLists.txt
@@ -82,7 +82,20 @@ add_library(webgpu_dawn SHARED IMPORTED)
 set_property(TARGET webgpu_dawn PROPERTY IMPORTED_LOCATION "${WEBGPU_LIBS_PATH}/libwebgpu_dawn.so")
 
 # Link
-target_link_libraries(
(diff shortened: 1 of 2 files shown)


wcandillon:
very cool thank you :)

PR #197, opened 2025-01-14
Update TypeGPU and use the new resolve API

reczkok:
This PR utilizes TypeGPU's new resolve API to link tgpu resources with shader code. It allows for a single type-safe source of truth for structs and automation for bindings. It also updates deprecated calls to `populate`.


wcandillon:
nice :) :) thank you for making these

wcandillon (approved):

PR #202, opened 2025-02-03
chore(test app, apple): disable metal api validation

Saadnajmi:
Disable Metal API Validation in the test app, required to unblock E2E tests with the test app.


wcandillon:
yes :) :) that's a big one thank you 😊

wcandillon:
this works great big thank you 🙏

PR #205, opened 2025-02-07
fix: resolve error in Three.js demo when enabling subgroup feature

MrQinYQ:
In some models, adapter.features has subgroup features, but the adapter.requestDevice parameter does not enumerate all supported features during conversion, causing the three.js demo to fail to run in some models.
This pr completes these enumerations.
image
image
image
image


wcandillon:
very good thanks :)
Also very cool that you are using threejs with RN WebGPU, I've been playing quite a bit with it this week 😇

PR #228, opened 2025-06-03
Add Web Support (Take 2)

scottmas:
Finally got around to adding vanilla web support without a companion Expo project. Just need some tweaks to the metro.config.js to make it support web as well. 

To test, go to the example app and do `yarn web`, then open http://localhost:8081

I was UNABLE to get MNISTInference example working fully - Nothing gets drawn to the Skia canvas, and the number 2 is always predicted. I'm honestly quite baffled on the lack of visual feedback, since `global.CanvasKit` is present and no errors are generated. And having 2 always be predicted maybe is an artifact of changing how the font is loaded? Sounds like something maybe the creator of react-native-skia should have a look at...

Let me know if there's any changes you'd like me to make!


wcandillon:
Thank you for doing this. There are a lot of changes which seem to be unrelated to the PR itself.
We could add an expo project to `apps` but I think expo project don't really support workspaces. But maybe if we use the project for web only (we don't use it for iOS or Android), it would work fine?
In Skia, we use remotion a lot so we have it in the apps/ and we also have an external-apps folder for apps disconnected from the monorepo: https://github.com/wcandillon/react-native-skia/tree/main/external-apps/expo-router-app

wcandillon:
is this "all" it takes to get the Web version working? :) it looks small enough :)

packages/webgpu/src/Canvas.tsx
```diff
@@ -11,50 +11,19 @@ import {
 } from "react";
 import type { RefObject } from "react";
```
wcandillon, line 12: shouldn't this PR only add `src/Canvas.web.tsx`,  and `src/main.web.ts` and that's it sorry if I am missing something.

apps/example/getWebMetroConfig.js
```diff
@@ -0,0 +1,34 @@
+const path = require("path");
+const root = path.resolve(__dirname, "../..");
+const r3fPath = path.resolve(root, "node_modules/@react-three/fiber");
```
wcandillon, line 3: this import is not used here right?

PR #231, opened 2025-06-21
Use singleton for WebGPU instance

wcandillon:
the idea is to be use that the same instance is always used even in case of hot reload and so on


wcandillon:
@Copilot can you fix the build errors in this PR?

PR #249, opened 2025-09-08
@arthwood/android transparent fix

arthwood:
It fixes the crash due to invalid, non-positive arguments passed to `ImageReader` new instance.


wcandillon:
Closing it for now. We fixed WebGPUAHBView.java issues but it doesn't work with resizing so it is disabled in the latest version (it was actually an oversight that this view type was enabled).

Issue #270, opened 2025-10-16
Crash using webgpu with other nitro modules


zispidd:
I have several nitro modules installed, and I started getting crashes periodically. Right now, the error is as follows:

<img width="1290" height="2796" alt="Image" src="https://github.com/user-attachments/assets/0a26408f-3038-418a-b6c8-f74970feb7c3" />

I asked @mrousavy about this, and he said the following:
<img width="690" height="383" alt="Image" src="https://github.com/user-attachments/assets/a6e65b29-234a-4c48-a2d5-e13ffeda0718" />

Can you please fix this @wcandillon? I can't use typegpu-confetti because of this. Thank you!

PR #271, opened 2025-10-17
fix: Rename `Dispatcher` global name to avoid conflicts with Nitro

mrousavy:
Since WebGPU uses an early copy of Nitro (kinda like a fork), there are some conflicts that users will run into when they install WebGPU in an app that already has Nitro Modules installed (or other way around).

An official full fix would be to migrate WebGPU to Nitro, not only for better maintenance, but also for better performance and stability. I really don't think the migration is a huge effort.

For now, this PR fixes the name of the `Dispatcher` that is injected into global, as Nitro uses the same name and when one of the two Nitro versions downcasts the global into the specific C++ symbol, we have conflics since it's two different C++ symbols - it's either `nitro::margelo::Dispatcher` (the official one) or the WebGPU `margelo::Dispatcher`.

It's essentially a race condition with no clear error log since its just a C++ pointer cast that fails (super hard to debug lol)

Fixes https://github.com/wcandillon/react-native-webgpu/issues/270

### DIFF
--- a/packages/webgpu/package.json
+++ b/packages/webgpu/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-native-wgpu",
-  "version": "0.2.9",
+  "version": "0.2.10",
   "description": "React Native WebGPU",
   "main": "lib/commonjs/index",
   "module": "lib/module/index",


wcandillon:
Thanks a lot!

PR #339, opened 2026-03-31
Fix the context lifecycle

Cyberax:
Sorry for the big PR. This is still a bit of work-in-progress, but it's now in a state where it doesn't crash my Android emulator and renders all the examples correctly.

The handling of onscreen/offscreen transitions was broken in all kinds of interesting ways:

1. On Android, the rendering surface can go away at any point from the
   point of view of the JavaScript thread. There's no way to control it.
2. On iOS/macOS the surfaces never go away, but  we can't resize them
   outside the main thread.

The previous commits add an example of a virtualized list that exercises
the surface lifecycle quite vigorously.

Oh, and to add to the fun: Dawn WebGPU device is not threadsafe. So it
easily crashes even if all we want to do is copy an offscreen texture
to the screen, while the JavaScript thread is rendering.

This commit adds a GPU-level lock to all `NativeObject`-based wrappers,
so it is automatically taken on any operation. And the UI-thread based
logic takes this lock manually. This serializes all the GPU operations.

And because the iOS and Android implementations have diverged quite a
bit, the `SurfaceInfo` is now an abstract class `SurfaceBridge` that has
platform-specific implementations.

### DIFF
--- a/apps/example/android/gradle.properties
+++ b/apps/example/android/gradle.properties
@@ -44,6 +44,7 @@ hermesEnabled=true
 # Note that this is incompatible with web debugging.
 #newArchEnabled=true
 bridgelessEnabled=true
+edgeToEdgeEnabled=true
 
 # Uncomment the line below to build React Native from source.
 #react.buildFromSource=true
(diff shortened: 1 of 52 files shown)


sindreij:
@Cyberax: Is your branch ready to be used?

We're in the process of moving a big part of our application over to this library, however making it work correctly on Android seems difficult. For instance, we have issues where when we navigate to a new page and then back, rendering suddenly don't work anymore.

Please let us know if there is anything we could do to help getting this fixed

PR #374, opened 2026-06-02
feat: add Dawn device toggles via dawnToggles

pepperoni505:
This PR adds the ability to set Dawn **device** (instance and adapter toggles aren't supported by this PR) toggles (full list [here]( https://dawn.googlesource.com/dawn/+/refs/heads/main/src/dawn/native/Toggles.cpp)).

Our use case is the `ignore_imported_ahardwarebuffer_vulkan_image_size toggle`: we hit validation errors in some Android emulator environments due to the driver issue documented in the file linked above. Rather than patch this repo to hardcode that one toggle, exposing the ability to set any toggle is the cleaner, more general solution.

```ts
const device = await adapter.requestDevice({
	dawnToggles: {
		enabledToggles: ["ignore_imported_ahardwarebuffer_vulkan_image_size"],
		// disabledToggles: [...],
	},
});
```

On the API shape: this adds a non-standard field to `GPUDeviceDescriptor` rather than touching the `requestDevice` signature. It's the most logical home, but I'm open to alternatives if you'd prefer it elsewhere.


wcandillon:
looks good!

wcandillon (approved):

PR #375, opened 2026-06-02
feat(🚧): add fence for shared texture memory

### LINKED ISSUE #369: SharedTextureMemory: expose fences on beginAccess/endAccess?
The `GPUSharedTextureMemory` header notes that fence-based sync isn't exposed yet:
https://github.com/wcandillon/react-native-webgpu/blob/0ccc35aa411a4e0f636f1fddab10ee6599af5f42/packages/webgpu/cpp/rnwgpu/api/GPUSharedTextureMemory.h#L34-L37

We have a use case that needs it: we import an app-owned IOSurface / AHardwareBuffer via `importSharedTextureMemory`, render into it with the WebGPU device, then composite that surface into a separate GPU pass (Mapbox) on a different queue, so we need the `MTLSharedEvent` Dawn exports from `EndAccess` on iOS, and the `sync-fd` on Android to protect against texture reads before rendering is finished.

I've prototyped this and am happy to open a PR, but wanted your take on the API shape first. The current sketch:
- `device.importSharedFence({ type, handle })` → `GPUSharedFence` (with `.export()`)
- `beginAccess(texture, initialized, fences?, signaledValues?)` - wait fences
- `endAccess(texture)` → `{ fences, signaledValues, ... }` - surfaces the fences (this changes the current `boolean` return)

Does this match what you had in mind, or would you prefer a different shape?

pepperoni505:
Closes #369 

Adds `GPUSharedFence`, exposing Dawn's shared-fence primitives so `GPUSharedTextureMemory` can be synchronized across queues/APIs. Until now `beginAccess`/`endAccess` only took the no-fence path, which is fine for single-producer still frames, but it can't safely coordinate an external producer and a GPU consumer on different queues (e.g. a video/camera surface being written while Dawn samples it).


wcandillon:
I like it a lot, thank you for the PR.

In my case I was flip flopping between 1. enabling people to do everything in JS using the SharedTextureMemory API or 2. pushing the importExternalTexture API.

And I realized that it's both 1. and 2.

Item 1 makes a lot of sense considering that this is React Native. You want to give full access/control over the native APIs. And I have a list of items there and this PR is part of that list, so again thank you for doing this.

And item 2 is also unavoidable anyway for zero-copy YUV texture support on Android.
So now my mental model on how to approach this is a bit tidier than yesterday.

packages/webgpu/src/__tests__/SharedTextureMemory.spec.ts
```diff
@@ -41,13 +41,7 @@ describe("SharedTextureMemory", () => {
             label: "test-frame",
           });
           const texture = memory.createTexture();
-          if (!memory.beginAccess(texture, true)) {
```
wcandillon, line 44: does this change make sense? Can you explain me?

apps/example/src/SharedTextureMemory/SharedTextureMemory.tsx
```diff
@@ -202,11 +202,7 @@ export const SharedTextureMemory = () => {
           label: "video-frame",
         });
         const texture = memory.createTexture();
-        if (!memory.beginAccess(texture, true)) {
-          texture.destroy();
-          frame.release();
-          return null;
-        }
+        memory.beginAccess(texture, true);
```
wcandillon, line 205: does this change make sense? Can you explain me?

wcandillon (changes requested):
Looks good! I am just confused about the two changes I commented on.

packages/webgpu/cpp/rnwgpu/api/GPUSharedFence.cpp
```diff
@@ -0,0 +1,72 @@
...
+  // caller-owned (the caller must close() it), so dup() it. Without the dup the same fd is closed twice —
+  // once by the caller and once by Dawn on fence destruction — tripping Android's fdsan (double-close abort).
+  wgpu::SharedFenceSyncFDExportInfo fdInfo{};
+  info.nextInChain = &fdInfo;
+  _instance.ExportInfo(&info);
+  int exportedFd =
+      fdInfo.handle >= 0 ? ::fcntl(fdInfo.handle, F_DUPFD_CLOEXEC, 0) : fdInfo.handle;
+  handle = static_cast<uint64_t>(static_cast<uint32_t>(exportedFd));
+#else
+  _instance.ExportInfo(&info);
```
wcandillon:
```suggestion
// react-native-webgpu only targets Apple (Metal) and Android (Vulkan). On any
// other platform there is no native handle convention to expose, so fail loudly
// rather than handing back a meaningless handle of 0.
throw jsi::JSError(runtime,
                    "GPUSharedFence::export(): unsupported platform (only "
                    "Apple/Metal and Android/Vulkan are supported)");
```

packages/webgpu/cpp/rnwgpu/api/GPUSharedFence.h
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  static void definePrototype(jsi::Runtime &runtime, jsi::Object &prototype) {
+    installGetter(runtime, prototype, "__brand", &GPUSharedFence::getBrand);
+    installMethod(runtime, prototype, "export", &GPUSharedFence::exportInfo);
+    installGetterSetter(runtime, prototype, "label", &GPUSharedFence::getLabel,
+                        &GPUSharedFence::setLabel);
+  }
+
+  inline const wgpu::SharedFence get() { return _instance; }
```
wcandillon:
```suggestion
  inline wgpu::SharedFence get() { return _instance; }
```

packages/webgpu/src/__tests__/SharedTextureMemory.spec.ts
pepperoni505: Sorry, I probably should've highlighted that better in the PR description. Given the change to `endAccess`'s return type (to return the fences), I felt it was a bit awkward to include a `success` field in that type as well. I looked at the rest of the codebase and noticed `beginAccess` and `endAccess` were the only functions to return a "success value", instead of throwing. So I changed both functions to throw, in order to keep consistency between them (technically `beginAccess` could've remained returning a boolean, but that wouldn't line up with `endAccess`).

wcandillon: got it thks

PR #376, opened 2026-06-02
chore(🧵): worklet-reachable native-buffer factory via RNWebGPU capture

wcandillon:


fax1ty:
Realtime MP4 encoding even on low end devices! Finally! 

https://github.com/user-attachments/assets/189740f8-3ffa-4712-a98e-2ac4cd746bec

wcandillon:
that's awesome :)

PR #418, opened 2026-07-15
fix(⏰): cache RuntimeContext by runtime address instead of runtimeData

reczkok:
## Summary
WebGPU crashes natively (EXC_BAD_ACCESS) when it's used from a worklet runtime, e.g. navigator.gpu.requestAdapter() or GPUBuffer.mapAsync() from the Reanimated UI thread. On the main JS runtime it works fine. Also all those accesses are via TypeGPU and I didn't test the raw WebGPU cases but I reckon it's the exact same

I'm not a Hermes/C++ expert, so i don't fully grasp why the crashes happened exactly but i gathered the .ips logs from crashes and run them through Fable which produced a fix that works for me locally

# Everything that follows (both the explanation and the fix is authored by AI and is attached just as a hypothesis and potential fix that worked for me, please take that into account while looking) - feel free to take over this branch or close this in favor of a better fix

Two on-device .ips crash reports, both from GPUBuffer.mapAsync called on the Reanimated UI runtime

```
Report 1, EXC_BAD_ACCESS (SIGKILL), KERN_PROTECTION_FAILURE at 0x0:
0  ???        0x0
1  hermesvm   HermesRuntimeImpl::setRuntimeDataImpl(jsi::UUID const&, ...) + 84
2  <app>      jsi::WithRuntimeDecorator<worklets::AroundLock, ...>::setRuntimeDataImpl(...)
3  <app>      rnwgpu::async::RuntimeContext::getOrCreate(jsi::Runtime&, wgpu::Instance) + 248
4  <app>      rnwgpu::GPUBuffer::mapAsync(jsi::Runtime&, ...) + 168
```

```
Report 2, EXC_BAD_ACCESS (SIGSEGV), KERN_INVALID_ADDRESS at 0x8e0008:
0  <app>      jsi::Runtime::getRuntimeData(jsi::UUID const&) + 32
1  <app>      rnwgpu::async::RuntimeContext::get(jsi::Runtime&) + 44
2  <app>      rnwgpu::async::RuntimeContext::getOrCreate(jsi::Runtime&, wgpu::Instance) + 40
3  <app>      rnwgpu::GPUBuffer::mapAsync(jsi::Runtime&, ...) + 168
   ...        worklets::WorkletRuntime::runSync(...) -> executeQueueForProMotion
```

Both fault inside Hermes' set/getRuntimeData on the way out of RuntimeContext::get/getOrCreate, and only on a worklet runtime. I disassembled the shipping hermesvm binary (Mach-O UUID matched the reports) and the faulting instructions read/delete an uninitialized bucket in the map Hermes keeps behind runtimeData.

### What

RuntimeContext cached itself on the runtime via jsi::Runtime::set/getRuntimeData with a default-constructed (all-zero) jsi::UUID as the key. Hermes seems to reserve that exact all-zero UUID as the "empty slot" marker in that map, so reading or writing under it hits uninitialized memory. On the main runtime WebGPU is the first thing to use runtimeData so it happens to work; on worklet runtimes (Reanimated seeds runtimeData first) the first WebGPU call reads or frees garbage, which is the two crashes above.

### Fix

Stop using runtimeData. Cache RuntimeContext in a small static, mutex-guarded map keyed by thmeRegistry.h). The runtime object stays put for its lifetime so its address is a safe key,entries are weak_ptrs (the GPU objects already own the context), and stale entries get pruned lazily. Sidesteps the UUID issue entirely and behaves the same across JS engines.

### DIFF
--- a/packages/webgpu/cpp/rnwgpu/async/RuntimeContext.cpp
+++ b/packages/webgpu/cpp/rnwgpu/async/RuntimeContext.cpp
@@ -7,14 +7,12 @@
 #include <ReactCommon/CallInvoker.h>
 
 #include "AsyncTaskHandle.h"
+#include "RuntimeRegistry.h"
 #include "WGPULogger.h"
 
 namespace rnwgpu::async {
 
 namespace {
-struct RuntimeData {
-  std::shared_ptr<RuntimeContext> context;
-};
 constexpr const char *TAG = "RuntimeContext";
 
 // The main JS runtime and its CallInvoker, registered once on install. The
@@ -31,6 +29,11 @@ std::mutex &processEventsMutex() {
   static std::mutex mutex;
   return mutex;
 }
+
+RuntimeRegistry<jsi::Runtime, RuntimeContext> &runtimeContexts() {
+  static RuntimeRegistry<jsi::Runtime, RuntimeContext> registry;
+  return registry;
+}
 } // namespace
 
 void RuntimeContext::registerMainRuntime(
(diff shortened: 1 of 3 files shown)


wcandillon:
https://github.com/wcandillon/react-native-webgpu/pull/422 should fix the issue but let's keep coordinate on this.

Issue #432, opened 2026-07-24
createImageBitmap and copyExternalImageToTexture do not preserve alpha semantics on Apple platforms


Josema:
I encountered this issue while implementing MTSDF atlases. Codex prepared a fix for me, but I don’t want to open a pull request because I’m not confident about the code or the patch. Here is what Codex suggests to fix the problem:

---

## Description

On Apple platforms, images decoded by `createImageBitmap` are always stored with premultiplied RGB channels.

Additionally:

- `createImageBitmap` ignores `ImageBitmapOptions.premultiplyAlpha`.
- `GPUQueue.copyExternalImageToTexture` ignores `GPUImageCopyTextureTagged.premultipliedAlpha`.
- The same behavior occurs when `flipY` is enabled.

This makes it impossible to upload straight/unpremultiplied RGBA data without losing information.

## Current behavior

The Apple implementation creates the bitmap using:

```objc
kCGImageAlphaPremultipliedLast | kCGBitmapByteOrder32Big
```

Core Graphics therefore transforms decoded pixels as follows:

```text
RGB_stored = RGB_source × alpha
```

For example, a source pixel containing:

```text
RGBA = [128, 128, 128, 128]
```

is decoded approximately as:

```text
RGBA = [64, 64, 64, 128]
```

The original straight RGB values are no longer available after quantization.

`copyExternalImageToTexture` then uploads these bytes directly through `WriteTexture`, regardless of the value of:

```ts
destination.premultipliedAlpha
```

### `createImageBitmap`

The native implementation should respect:

```ts
createImageBitmap(source, {
  premultiplyAlpha: "none"
})
```

When `"none"` is requested, the resulting `ImageBitmap` should preserve straight RGBA values:

```text
[128, 128, 128, 128]
```

When `"premultiply"` is requested, the resulting values should be premultiplied:

```text
[64, 64, 64, 128]
```

The bitmap should retain enough metadata to identify whether its stored data is premultiplied.

### `copyExternalImageToTexture`

`GPUQueue.copyExternalImageToTexture` should convert between the source representation and the representation requested by:

```ts
destination.premultipliedAlpha
```

The default value of `GPUImageCopyTextureTagged.premultipliedAlpha` must remain `false`.

The conversion should happen exactly once:

| Source representation | Destination value | Expected operation |
| --- | ---: | --- |
| Straight | `false` or omitted | No conversion |
| Straight | `true` | Premultiply |
[...]

PR #433, opened 2026-07-24
Preserve straight-alpha ImageBitmap data on Apple platforms

Josema:
https://github.com/wcandillon/react-native-webgpu/issues/432


wcandillon:
@Josema Thank You for this. I updated the test suite and ran it against google chrome to generate the exact expected results. Can you run it on Android and iOS and fix the remaining issues if any?

Josema:
@wcandillon your implementation (https://github.com/wcandillon/react-native-webgpu/pull/434) works perfect on my side. Thanks a lot for the quick fix and this amazing library.

PR #435, opened 2026-07-27
fix: respect source.origin in copyExternalImageToTexture

Josema:
## Problem

The source descriptor already parsed and stored `origin`, but `copyExternalImageToTexture()` always uploaded pixels starting at the beginning of the `ImageBitmap` backing store.

For example, given this 2×2 image:

```text
A B
C D
```

Copying the right column:

```ts
device.queue.copyExternalImageToTexture(
  {
    source: bitmap,
    origin: [1, 0],
  },
  { texture },
  {
    width: 1,
    height: 2,
  },
);
```

should upload:

```text
B
D
```

Instead, it uploaded:

```text
A
C
```

This is especially visible when using texture atlases. Copies intended to generate padding from the last row or column could read pixels from the beginning of the image, producing a solid line along the right or bottom edge.

The existing `flipY` path had a related issue: it flipped the complete bitmap before applying the copy extent. For a subregion, WebGPU expects only the selected region to be flipped.

## Solution

The fix is intentionally limited to the existing `copyExternalImageToTexture()` implementation:

- Read `source.origin`, defaulting to `[0, 0]`.
- Validate that `origin + copySize` fits inside the source bitmap.
- Use the staging path when the origin is non-zero, or when `flipY` or alpha
  conversion is required.
- Copy only the selected rows and columns into the staging buffer.
- Reverse the selected rows while copying when `flipY` is enabled.
- Apply alpha conversion to the staged subregion when required.
- Upload the staging buffer with row layout matching the subregion.

The existing direct upload path remains unchanged for zero-origin copies that
do not require `flipY` or alpha conversion.

## Validation

The regression suite covers:

- A non-zero horizontal origin.
- A non-zero vertical origin.
- `source.origin` combined with alpha conversion.
- `source.origin` combined with `flipY`.
- Horizontal and vertical out-of-bounds source regions.
- The existing zero-origin path.

Verified with:

```bash
yarn workspace react-native-webgpu test:ref \
  --runTestsByPath src/__tests__/ImageBitmapOrigin.spec.ts

yarn workspace react-native-webgpu test:node \
  --runTestsByPath src/__tests__/ImageBitmapOrigin.spec.ts

yarn workspace react-native-webgpu test \
  --runTestsByPath src/__tests__/ImageBitmapOrigin.spec.ts
```

Results:

- Chrome reference client: 7/7 passing.
- `dawn.node`: 7/7 passing.
- iOS Simulator (iPhone 11, iOS 26.2): 7/7 passing.
- TypeScript and ESLint checks passing.
- Native C++ source compiled successfully.

### DIFF
--- a/packages/webgpu/cpp/rnwgpu/api/GPUQueue.cpp
+++ b/packages/webgpu/cpp/rnwgpu/api/GPUQueue.cpp
@@ -129,8 +129,20 @@ void GPUQueue::copyExternalImageToTexture(
   }
 
   const auto origin = source->origin.value_or(nullptr);
-  const size_t sourceOriginX = origin ? origin->x : 0;
-  const size_t sourceOriginY = origin ? origin->y : 0;
+  // GPUOrigin2D coordinates are [EnforceRange] unsigned long: negative,
+  // NaN, or out-of-range doubles must be rejected here because casting
+  // them to an unsigned integer is undefined behavior.
+  constexpr double kMaxOrigin =
+      static_cast<double>(std::numeric_limits<uint32_t>::max());
+  if (origin && (!(origin->x >= 0) || !(origin->y >= 0) ||
+                 origin->x > kMaxOrigin || origin->y > kMaxOrigin)) {
+    throw std::runtime_error(
+        "The source origin must be a non-negative integer coordinate.");
+  }
+  const size_t sourceOriginX =
+      origin ? static_cast<size_t>(origin->x) : 0;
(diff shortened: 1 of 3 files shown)


wcandillon:
@Josema I see it marked as draft, let me know once it is ready for review 👍

Josema:
Ready. Again, this is a fix from Codex that works for me. Feel free to close the PR if you don't think the fix is good enough.

wcandillon (approved):
Thanks 👍

Issue #444, opened 2026-08-14
Feature request: expose the surface present mode (FIFO is hardcoded in two places) and request the panel refresh rate


smm-h:
Hi — first, thank you for this library. I have a React Native app rendering a three.js WebGPU scene through it on Dawn/Vulkan, and it works; this is a request from having profiled that app on a device, not a complaint.

Two related things about frame delivery, with the evidence I gathered.

## 1. Present mode is hardcoded to FIFO, in two places

Verified against the installed `react-native-webgpu@0.8.2` and re-checked against `main` (same line numbers in both):

- `packages/webgpu/cpp/rnwgpu/api/GPUCanvasContext.cpp:30`

  ```cpp
  surfaceConfiguration.presentMode = wgpu::PresentMode::Fifo;
  ```

  set unconditionally at the end of `GPUCanvasContext::configure()`, after the format/usage conversion.

- `packages/webgpu/cpp/rnwgpu/SurfaceRegistry.h:242`

  ```cpp
  _config.presentMode = wgpu::PresentMode::Fifo;
  ```

  inside `SurfaceInfo::configure()`, which **re-stamps** FIFO over whatever `wgpu::SurfaceConfiguration` the caller passed in, right after the width/height clamp.

The second one is what makes this a hard wall rather than a default: even if a JS caller could express a present mode, `SurfaceRegistry::configure()` would overwrite it. And there is no way to express one anyway — `GPUCanvasConfiguration` (`packages/webgpu/cpp/rnwgpu/api/descriptors/GPUCanvasConfiguration.h:16`) carries only `device`, `format`, `usage`, `viewFormats` and `alphaMode`. `wgpu::PresentMode` itself has `FifoRelaxed`, `Immediate` and `Mailbox` available (`cpp/webgpu/webgpu_cpp.h:505-511`), so the Dawn side already supports the alternatives.

I understand why FIFO is the safe default (it is the only mode guaranteed present on every Vulkan surface, and it is what the web canvas contract implies). The ask is not to change the default — it is to make it reachable.

### Why this shows up as a frame-rate cliff

FIFO is a strict queue, and Dawn's Vulkan backend configures the surface with a minimal image count for it. With a shallow chain there is essentially no pipelining slack: the app cannot get ahead of the compositor, so `getCurrentTexture()` blocks the JS thread waiting for an image, and a frame that misses one vsync waits for the whole next one.

[...]

PR #453, opened 2026-08-25
feat: expose native canvas present modes

huytdps13400:
## Summary

- add a non-standard `presentMode` field to `GPUCanvasConfiguration`
- map `fifo`, `fifo-relaxed`, `immediate`, and `mailbox` to Dawn's native surface modes
- keep FIFO as the default and stop `SurfaceRegistry` from overwriting the caller's selection
- document platform-dependent support and add TypeScript/native regression coverage

This addresses the present-mode portion of #444. It intentionally does not add DisplayLink/Choreographer ownership or a refresh-rate scheduler, matching the maintainer's scope guidance in the issue.

Unsupported modes remain subject to Dawn's existing surface configuration validation.

## Verification

- `yarn workspace react-native-webgpu tsc`
- `yarn workspace react-native-webgpu test:node CanvasConfiguration.spec.ts` (3 tests passed)
- `yarn workspace react-native-webgpu build`
- `yarn workspace docs tsc`
- `yarn workspace docs build:docs`
- `clang++ -std=c++20 -fsyntax-only ... GPUCanvasContext.cpp`
- `cpplint` on all changed C++ files

The full local Node-WebGPU suite also ran; all relevant tests passed, with one unrelated existing Node backend failure in `GPU.spec.ts` where `limits.maxBindGroups` serializes as `undefined`.


wcandillon:
Can you give me a motivating example to expose this feature? I think it could make sense but I would like to understand the use-cases

huytdps13400:
> Can you give me a motivating example to expose this feature? I think it could make sense but I would like to understand the use-cases
Thanks, that makes sense. I’ll prepare a motivating example and some concrete use‑cases for ‎`presentMode` and get back to you.

huytdps13400:
A concrete motivating case is a touch-driven 3D editor or drawing canvas where a gesture/worklet continuously updates the camera or brush state and each render represents the latest input.

With FIFO, pending frames are presented in submission order. If the app has queued work, the screen can show an older camera/brush state even though newer input has already been rendered. Mailbox still waits for vblank (so no tearing), but replaces the single pending frame with the newest one. That trades some power for lower input-to-photon latency, which is exactly the use case the Vulkan performance guide calls out for mailbox mode.

For example, an app could choose mailbox only while the user is actively orbiting a model or drawing, then reconfigure back to FIFO when interaction ends:

```ts
context.configure({
  device,
  format: navigator.gpu.getPreferredCanvasFormat(),
  presentMode: isInteracting ? "mailbox" : "fifo",
});
```

`fifo-relaxed` is useful for a mostly-vsynced loop that occasionally finishes just after vblank; `immediate` is mainly an opt-in profiling/latency escape hatch for apps willing to accept tearing. FIFO remains the default and the best general-purpose/power-efficient choice.

The Vulkan definitions are here: https://registry.khronos.org/vulkan/specs/latest/man/html/VkPresentModeKHR.html. Khronos also documents mailbox specifically as a lower-input-latency option, with the power tradeoff: https://docs.vulkan.org/samples/latest/samples/performance/swapchain_images/README.html. Dawn already rejects a mode not reported by the surface capabilities, so this PR exposes the choice without promising every mode on every device.

The main product motivation is therefore mailbox. I mapped all four Dawn modes for completeness, but I can narrow the public extension to `fifo`/`mailbox` if you would prefer the smallest API surface.

wcandillon:
Could you add it as an example to this PR? If it's Android only, that's ok for me. I just want to get a better sense of this.

huytdps13400:
Added the requested Android example in e466e18.

The new Present Modes screen:
- toggles the canvas between fifo and mailbox
- drives both a white Reanimated UI marker and a cyan WebGPU marker from the same pan gesture, so the input-to-presentation behavior can be compared directly
- keeps the current gesture position while switching modes
- stops the previous render loop before presenting again after a mode change

Verification:
- yarn workspace Example tsc
- targeted ESLint for PresentMode.tsx, App.tsx, Home.tsx, and Route.ts
- Android debug build, install, and launch on Pixel 10 Pro emulator (API 37)
- runtime QA: FIFO and MAILBOX both rendered, mode switching preserved marker position, and drag updates stayed aligned
- no Dawn validation errors or crashes in the native log

wcandillon:
this is compelling thank you

apps/docs/content/docs/getting-started/canvas.mdx
```diff
@@ -61,6 +61,24 @@ context.present(); // React Native only
 
 `present()` is a React Native only method (it is not part of the Web WebGPU spec). It runs **synchronously** on the calling thread, so the frame is presented from whichever thread did the rendering. This works the same on every runtime: the main JS runtime, the UI thread, and dedicated worklet runtimes. Always call it after `submit()`, never before.
 
+### Present mode
```
wcandillon, line 64: On iOS, mailbox and immediate are no-ops (Dawn's Metal swapchain only toggles vsync on macOS), fifo-relaxed always fails, and the validation error only reaches the uncaptured-error callback, never a configure() throw. The docs should describe per-backend behavior and note the example is Android-only.

packages/webgpu/cpp/rnwgpu/api/descriptors/GPUCanvasConfiguration.h
```diff
@@ -63,6 +64,23 @@ struct JSIConverter<std::shared_ptr<rnwgpu::GPUCanvasConfiguration>> {
           result->alphaMode = wgpu::CompositeAlphaMode::Premultiplied;
         }
       }
+      if (value.hasProperty(runtime, "presentMode")) {
```
wcandillon, line 67: Explicit undefined presentMode throws (GPUCanvasConfiguration.h:68). The code uses hasProperty plus asString, so presentMode: cond ? "mailbox" : undefined throws a raw JSI error instead of applying the FIFO default. Guard with isUndefined/isString before reading, like the other optional members in that file.

packages/webgpu/cpp/rnwgpu/api/GPUCanvasContext.cpp
```diff
@@ -27,7 +27,7 @@ void GPUCanvasContext::configure(
 #ifdef __APPLE__
   surfaceConfiguration.alphaMode = configuration->alphaMode;
 #endif
-  surfaceConfiguration.presentMode = wgpu::PresentMode::Fifo;
+  surfaceConfiguration.presentMode = configuration->presentMode;
```
wcandillon, line 30: Unsupported presentMode fails silently

apps/example/src/PresentMode.tsx
```diff
@@ -0,0 +1,292 @@
...
+      context,
+      navigator.gpu.getPreferredCanvasFormat(),
+      mode,
+      canvas.width,
+      canvas.height,
+      pixelRatio,
+    );
+
+    return () => {
+      running.value = false;
```
wcandillon, line 152: Render loop leaks on effect re-run . Cleanup sets a shared boolean to false, but the immediate re-run sets it back to true before the UI thread reads it, so Fast Refresh spawns a second loop on the same context. Use a per-effect generation token.

packages/webgpu/src/__tests__/CanvasConfiguration.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    expect(configuration).toBeDefined();
+  });
+
+  it("configures a native surface with an explicit FIFO present mode", async () => {
+    if (client.OS !== "ios" && client.OS !== "android") {
+      return;
+    }
+
+    const size = await client.eval(({ ctx, device, gpu }) => {
```
wcandillon, line 37: this doesn't really work, can you update the test suite to really use an image snapshot test?

wcandillon (changes requested):
The motivational example you provided is compelling thank you for that. I left some review comments

PR #456, opened 2026-08-27
Fix/implicit device synchronization

### LINKED ISSUE #455: A GPUDevice shared between JS runtimes crashes within seconds (iOS and Android)
## What happens

Render a canvas from the Reanimated UI runtime, then touch the same `GPUDevice` from the React JS runtime, and the app dies in 4 to 12 seconds. No exotic API needed. `queue.writeBuffer`, `queue.submit`, `device.createBuffer`, `buffer.mapAsync` and `context.present` each do it.

Five signatures so far:

- iOS, Metal assert in `-[_MTLCommandEncoder dealloc]`
- iOS, Metal assert `encodeSignalEvent:value:` during `Submit`
- iOS, Metal assert in `-[_MTLCommandBuffer commit]`, reached from `Instance::ProcessEvents`
- iOS, `AddressSanitizer: heap-use-after-free` in Dawn's `RingBufferAllocator`
- Android, `SIGSEGV` on the JS thread inside `BufferBase::MapAtCreation`

I chased this on iOS first and assumed it was a Metal quirk. It is not. Android/Vulkan crashes too, and so does a Release build with no sanitizer linked, so the Metal asserts are not just validation noise.

## Repro

Branch: https://github.com/AlirezaHadjar/react-native-webgpu/tree/repro/cross-runtime-device
It sits on `main` at 244ad53 and only touches `apps/example`.

It adds a screen under **Diagnostics -> "Cross-Runtime Device Stress"**. The UI runtime renders a canvas through `runOnUI` while the React JS runtime hammers the same device from a `setTimeout` pump. Each mode picks which API family both sides hit.

1. `yarn build:ios` or `yarn build:android` in `apps/example`
2. Home, scroll to the bottom, tap **Tests**, then **Cross-Runtime Device Stress**
3. It starts in `write` mode and dies in about 10 seconds. Tap `submit`, `map` or `present` for the other signatures.

On iOS you can pick the mode at launch:

```
xcrun simctl launch <udid> <bundle-id> -RNWebGPUCrossRuntimeMode map
```

Watch the `ui:` and `js:` counters under the title. If either stops climbing, that runtime died and nothing is racing any more, so "it survived" means nothing.

## What I measured

iPhone 17 Pro simulator, iOS 26.5, Debug with ASan, 30 seconds per mode:

| mode | result | crash |
|---|---|---|
| write | crash 12s | `-[_MTLCommandEncoder dealloc]` <- `EnsureBlit` <- `QueueBase::WriteBuffer` |
| submit | crash 6s | ASan + `encodeSignalEvent` <- `Submit` |
| create | survived | |
| map | crash 4s | `-[_MTLCommandBuffer commit]` <- `SubmitPendingCommandBuffer` <- `DeviceBase::Tick` <- `Instance::ProcessEvents` |
| present | crash 4s | ASan + `encodeSignalEvent` <- `Submit` |
| churn | survived | |
| destroy | survived | |
| all | crash 6s | ASan + `Submit` |

Release build for the simulator, no sanitizer in the image list: `write` crashes in about 10 seconds with the same `-[_MTLCommandEncoder dealloc]` stack.

Android emulator, API 36, arm64, Debug: `submit` crashes with `SIGSEGV` on the React JS thread inside `createBuffer`.

## Stacks

iOS, both runtimes sitting in `rnwgpu::GPUQueue::writeBuffer` at the same time. Which thread aborts changes run to run:

```
abort -> __assert_rtn -> MTLReportFailure
-[_MTLCommandEncoder dealloc]
-[MTLSerializerBlitCommandEncoder dealloc]
dawn::native::metal::CommandRecordingContext::EnsureBlit()
dawn::native::metal::Device::CopyFromStagingToBuffer(...)
dawn::native::BufferBase::UploadData(...)
dawn::native::QueueBase::WriteBuffer(...)
wgpu::Queue::WriteBuffer(...)
rnwgpu::GPUQueue::writeBuffer(...)
```

iOS, the uploader's ring buffer torn up by one thread while another walks it:

```
AddressSanitizer: heap-use-after-free on address 0x0001047986c0
#1 dawn::SerialQueue<..., RingBufferAllocator::Request>::Enqueue(...)
#2 dawn::native::RingBufferAllocator::Allocate(...)
#3 dawn::native::DynamicUploader::Reserve(...)
   -> BufferBase::UploadData -> QueueBase::WriteBuffer -> rnwgpu::GPUQueue::writeBuffer
freed by thread T7 here: same Enqueue frame
previously allocated by thread T7 here: same Enqueue frame
T7 created by -[RCTJSThreadManager startJSThread] (RCTJSThreadManager.mm:69)
```

iOS `map` mode. The JS thread aborts inside the event pump while the main thread encodes:

```
thread com.facebook.react.runtime.JavaScript (triggered)
  -[_MTLCommandBuffer commit] <- metal::Queue::SubmitPendingCommandBuffer
  <- metal::Device::TickImpl <- DeviceBase::Tick <- InstanceBase::ProcessEvents
  <- rnwgpu::async::RuntimeContext::tick()
main thread
  -[MTLSerializerCommandEncoder endEncoding]
```

Android `submit` mode, tombstone on the React JS thread:

```
signal 11 (SIGSEGV), code 2 (SEGV_ACCERR), fault addr 0x76954777a8
tid: 10325, name: mqt_v_js
  #01 libwebgpu_dawn.so  dawn::native::BufferBase::MapAtCreation()+408
  #02 libwebgpu_dawn.so  dawn::native::DeviceBase::CreateBuffer(...)+368
  #03 libwebgpu_dawn.so  dawn::native::BufferBase::MapAtCreationInternal()+196
  #05 libwebgpu_dawn.so  dawn::native::DeviceBase::APICreateBuffer(...)+864
  #07 libreact-native-webgpu.so  rnwgpu::GPUDevice::createBuffer(...)+260
```

## Environment

- react-native-webgpu `main` at 244ad53
- react-native 0.81.4, react-native-reanimated 4.3.1
- iPhone 17 Pro simulator, iOS 26.5, Debug with ASan and Release without
- Android emulator, API 36, arm64, Debug

AlirezaHadjar:
Fixes #455.

## Why it crashes

`GPUAdapter::requestDevice` never asks for Dawn's `ImplicitDeviceSynchronization`, so `DeviceBase::mMutex` stays null (`Device.cpp:449`) and every `device->GetGuard()` inside Dawn turns into a no-op. Meanwhile rnwgpu hands the same device to every runtime it is boxed into (`installWebGPU`, GPUQueue through the Worklets serializer) and drives it from `SurfaceInfo` as well. Two threads then walk Dawn's queue, uploader and tick state at once, and Metal or Vulkan falls over a few seconds later.

## The change

`requestDevice` requests the feature when the adapter advertises it. Dawn's generated entry points then take `device->GetGuard()` (see the `ProcTable.cpp` template), and the paths a hand-rolled queue lock would miss guard themselves: `DeviceBase::APITick` (`Device.cpp:1589`), `Surface::Configure` / `GetCurrentTexture` / `Present` (`Surface.cpp:495,577,634`), and `metal::WaitForCommandsToBeScheduled` (`MetalBackend.mm:41`). Command encoding stays unlocked by Dawn's design, which matches one encoder per thread.

This is the supported way to use one device from several threads, and the same mechanism Chrome relies on. Single-runtime apps pay an uncontended mutex per call. Two runtimes serialize exactly when they would otherwise corrupt each other.

`GPUDevice::getFeatures` hides the feature when rnwgpu added it, so `device.features` keeps mirroring what the caller asked for. An explicit request still shows up, which is what `FeatureNames.spec.ts` asserts.

The other two commits add the diagnostic screen from the issue, under Diagnostics -> "Cross-Runtime Device Stress".

## Verification

iPhone 17 Pro simulator, iOS 26.5, Debug with ASan, 30 seconds per mode:

| mode | before | after |
|---|---|---|
| write | crash 12s | pass |
| submit | crash 6s | pass |
| create | pass | pass |
| map | crash 4s | pass |
| present | crash 4s | pass |
| churn | pass | pass |
| destroy | pass | pass |
| all | crash 6s | pass |

- 180 second soaks on `all` and `map`, frame and tick counters still climbing, so nothing deadlocks.
- Android emulator (API 36, arm64): `submit` crashed on the JS thread in `createBuffer` before, all seven modes run clean after.
- `yarn test` (jest e2e against the running app): 174/174.
- Reanimated examples (UI Thread, Dedicated Thread, Async Buffer UI) render as before.
- Throughput over 45 seconds of `write` is unchanged: ui 2971 / js 2940, against ui 2923 / js 2880 for a hand-rolled per-queue lock I benchmarked against.

## Alternative

A per-native-queue `std::mutex` in rnwgpu around the GPUQueue and SurfaceInfo entry points fixes 7 of the 8 modes. `map` still aborts, because `RuntimeContext::tick()` calls `Instance::ProcessEvents()` from every runtime and `DeviceBase::APITick` relies on `GetGuard()`. Covering that by hand means repeating the lock at every Dawn entry point two runtimes can reach.

### DIFF
--- a/apps/example/src/App.tsx
+++ b/apps/example/src/App.tsx
@@ -48,6 +48,7 @@ import { SharedTextureMemory } from "./SharedTextureMemory";
 import { ImportExternalTexture } from "./ImportExternalTexture";
 import { VisionCamera } from "./VisionCamera";
 import { SixteenBitTextures } from "./SixteenBitTextures";
+import { CrossRuntimeDevice } from "./Diagnostics/CrossRuntimeDevice";
 
 // The two lines below are needed by three.js
 import "fast-text-encoding";
@@ -126,6 +127,10 @@ function App() {
           />
           <Stack.Screen name="BackgroundDetach" component={BackgroundDetach} />
           <Stack.Screen name="SurfaceChurn" component={SurfaceChurn} />
+          <Stack.Screen
+            name="CrossRuntimeDevice"
+            component={CrossRuntimeDevice}
+          />
           <Stack.Screen
             name="StorageBufferVertices"
             component={StorageBufferVertices}
--- a/apps/example/src/Diagnostics/CrossRuntimeDevice.tsx
+++ b/apps/example/src/Diagnostics/CrossRuntimeDevice.tsx
@@ -1,0 +1,404 @@
+import React, { useCallback, useEffect, useRef, useState } from "react";
+import {
+  Button,
+  Platform,
+  ScrollView,
+  Settings,
+  StyleSheet,
+  Text,
+  View,
+} from "react-native";
+import type { CanvasRef, RNCanvasContext } from "react-native-webgpu";
+import { Canvas, installWebGPU } from "react-native-webgpu";
+import { runOnUI, useSharedValue } from "react-native-reanimated";
+import type { SharedValue } from "react-native-reanimated";
+
+const MODES = [
+  "write",
+  "submit",
+  "create",
+  "map",
+  "present",
+  "churn",
+  "destroy",
+  "all",
+] as const;
+
+type ReproMode = (typeof MODES)[number];
+
+const MODE_SETTING = "RNWebGPUCrossRuntimeMode";
+
+const OPS_PER_FRAME = 32;
+const OPS_PER_TICK = 32;
+const SURFACE_CHURN_INTERVAL_MS = 150;
+const DESTROY_AFTER_MS = 2_000;
+
+interface Shared {
+  device: GPUDevice;
+  format: GPUTextureFormat;
+  uiBuffer: GPUBuffer;
+  jsBuffer: GPUBuffer;
+}
+
+const readInitialMode = (): ReproMode => {
+  if (Platform.OS !== "ios") {
+    return "write";
+  }
+  const stored = Settings.get(MODE_SETTING);
+  return MODES.includes(stored) ? (stored as ReproMode) : "write";
+};
+
+const renderOnUI = (
+  running: SharedValue<boolean>,
+  uiFrames: SharedValue<number>,
+  device: GPUDevice,
+  context: RNCanvasContext,
+  buffer: GPUBuffer,
+  format: GPUTextureFormat,
+  mode: ReproMode,
+  ops: number,
+) => {
+  "worklet";
+
+  installWebGPU();
+  context.configure({ device, format, alphaMode: "premultiplied" });
+
+  const data = new Float32Array(4);
+  let frameNumber = 0;
+
+  const stress = () => {
+    if (mode === "write" || mode === "all") {
+      for (let i = 0; i < ops; i += 1) {
+        data[1] = i;
+        device.queue.writeBuffer(buffer, 0, data);
+      }
+    }
+    if (mode === "submit" || mode === "all") {
+      for (let i = 0; i < ops; i += 1) {
+        device.queue.submit([device.createCommandEncoder().finish()]);
+      }
+    }
+    if (mode === "create" || mode === "all") {
+      for (let i = 0; i < ops; i += 1) {
+        const scratch = device.createBuffer({
+          size: 256,
+          usage: GPUBufferUsage.COPY_DST | GPUBufferUsage.UNIFORM,
+        });
+        scratch.destroy();
+      }
+    }
+  };
+
+  const frame = () => {
+    if (!running.value) {
+      return;
+    }
+    frameNumber += 1;
+    uiFrames.value = frameNumber;
+    data[0] = frameNumber;
+    stress();
+
+    const encoder = device.createCommandEncoder();
+    const pass = encoder.beginRenderPass({
+      colorAttachments: [
+        {
+          view: context.getCurrentTexture().createView(),
+          clearValue: [
+            0.5 + 0.5 * Math.sin(frameNumber / 29),
+            0.5 + 0.5 * Math.sin(frameNumber / 31 + 2),
+            0.5 + 0.5 * Math.sin(frameNumber / 37 + 4),
+            1,
+          ],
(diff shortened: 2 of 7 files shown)


wcandillon:
Can you tell me what the stress test is doing? I am struggling to understand it? Can you tell me more at a high level?

AlirezaHadjar:
> Can you tell me what the stress test is doing? I am struggling to understand it? Can you tell me more at a high level?

[react-native-fast-confetti](https://github.com/AlirezaHadjar/react-native-fast-confetti) renderer depends on sharing one webgpu device across runtimes but rn webgpu previously exposed that supported pattern without enabling Dawn’s required thread synchronization. #456 makes the architecture safe by default.

AlirezaHadjar:
basically there's a very random race condition that is easier to trigger when stress test it

wcandillon:
why you don't do everything on the UI Thread? you can create the device on the JS thread (or even UI thread) and then pass it and use it only on the UI thread. Can you describe the concurrent patterns that you are using there?
I am trying to understand the use-case base.

AlirezaHadjar:
> why you don't do everything on the UI Thread? you can create the device on the JS thread (or even UI thread) and then pass it and use it only on the UI thread. Can you describe the concurrent patterns that you are using there? I am trying to understand the use-case base.

correct I just tried moving all operations on ui thread and couldn't repro the crash anymore. lmk if you think merging the pr would be useful or not since rn webgpu allows devices to cross runtimes and the race condition issue is still problem

wcandillon:
if the PR is to have implicit device synchronization option on, I would be ok with it but the current implementation in the PR seems a bit too complex for it, could it be simplified? (unless I'm missing something). The example looks a bit too convoluted as well.
And this should be documented as well (part of the native APIs section).

wcandillon:
Thank you for this, superseeding it with #467 467

PR #459, opened 2026-08-27
fix(ios): run surface transitions on main thread

### LINKED ISSUE #458: iOS: surface configuration mutates CAMetalLayer off the main thread
## Summary

On iOS, canvas surface attach and configuration can reach Dawn and mutate the backing `CAMetalLayer` from a React JavaScript or worklet runtime instead of the main thread. Core Animation layer mutation is main-thread-only, so the current behavior is undefined during lifecycle churn.

## Reproduction

**Reproduction branch:** [AlirezaHadjar/react-native-webgpu `repro/cametal-main-thread`](https://github.com/AlirezaHadjar/react-native-webgpu/tree/repro/cametal-main-thread)

This branch starts directly from upstream `main`. It changes only three files: it reuses the existing **Diagnostics -> Surface Churn** screen, opens and starts it automatically in iOS Debug, and adds a Debug-only assertion at the `CAMetalLayer` mutation site. It contains no fix.

1. Check out `repro/cametal-main-thread`.
2. Build and run `apps/example` as Debug on a physical iPhone.
3. The app opens **Surface Churn** and begins remounting and configuring the canvas every 400 ms.
4. Xcode prints:

```text
[cametal-repro] CAMetalLayer main-thread check: FAIL
```

The Debug assertion then stops at:

```text
CAMetalLayer mutations must run on the main thread
```

## Why it happens

Pending surfaces are adopted at frame boundaries on whichever runtime renders. The inactive-surface flush also uses the React JavaScript `CallInvoker`:

```cpp
_jsCallInvoker->invokeAsync(
    [info = std::move(info)] { info->applyPendingAttach(); });
```

Both `SurfaceInfo::configure()` and `applyPendingAttach()` can call `wgpu::Surface::Configure()`. On Metal, that operation configures the backing layer. `applyPendingAttach()` also assigns `CAMetalLayer.colorspace`.

## Physical iPhone reproduction data

This test targets issue #458 on hardware. No simulator measurements are included. I recorded five Xcode Activity Monitor runs on an iPhone 16 Pro Max running iOS 26.6.1, in this order: unfixed, unfixed, fixed, fixed, unfixed. Every run used the same Debug configuration, dependencies, signing, 15-second warmup, 60-second capture, fresh process, and 400 ms Surface Churn remount interval.

| Metric | Unfixed (upstream `main`) | Fix (#459) |
| --- | ---: | ---: |
| Runs | 3 | 2 |
| Total remounts | 450 | 300 |
| Mean app CPU | 31.78% | 16.64% |
| Average starting physical footprint | 378.52 MiB | 358.68 MiB |
| Average peak physical footprint | 388.97 MiB | 364.85 MiB |
| Average memory-growth slope | 7.06 MiB/min | 7.60 MiB/min |
| Crashes or WebGPU validation errors | 0 | 0 |

The Debug assertion on the unfixed reproduction branch reports `CAMetalLayer main-thread check: FAIL`, which reproduces the thread-confinement violation on a physical iPhone. During the same lifecycle workload, the fixed build used 47.7% less mean app CPU and about 20 to 24 MiB less physical footprint. The memory-growth slopes were effectively unchanged, so these runs do not establish a memory-leak fix.

## Proposed fix

- Run Apple surface attach, configure, reconfigure, unconfigure, and stale-surface recovery on the main thread.
- Send the inactive-surface attach flush directly to the main queue on Apple platforms.
- Dispatch before taking `SurfaceInfo`'s mutex, and release the mutex before synchronous stale-surface recovery to avoid lock inversion.
- Revalidate the frame epoch after dispatch so a lifecycle or frame change cannot be overwritten.
- Keep frame rendering, command encoding, queue submission, and presentation on their existing runtime.
- Add a Debug-only main-thread assertion at the `CAMetalLayer` mutation site.

## Acceptance criteria

- Every Apple operation that reads or mutates `CAMetalLayer` state runs on the main thread.
- Surface churn completes without deadlock or stale presentation.
- Rendering and Dawn queue work remain on their existing runtime.
- The physical-device Debug build passes, along with example lint and TypeScript checks.

AlirezaHadjar:
Fixes #458.

Moves Apple surface transitions to the main thread and retains detached surface state until the frame that acquired it reaches present. This fixes Core Animation thread confinement and prevents teardown from racing `getCurrentTexture()` / `queue.submit()`.

Validated on an iPhone 16 Pro Max with five 60-second Xcode Activity Monitor runs and 750 total remounts:

- mean app CPU: 31.78% to 16.64% (-47.7%)
- physical footprint: about 20 to 24 MiB lower

### DIFF
--- a/packages/webgpu/apple/MetalLayerColorSpace.mm
+++ b/packages/webgpu/apple/MetalLayerColorSpace.mm
@@ -1,7 +1,7 @@
 #import <CoreGraphics/CoreGraphics.h>
 #import <Foundation/Foundation.h>
 #import <QuartzCore/CAMetalLayer.h>
-#import <QuartzCore/CATransaction.h>
+#import <dispatch/dispatch.h>
 
 #include "webgpu/webgpu_cpp.h"
 
@@ -16,6 +16,15 @@ namespace rnwgpu {
 // behavior: identical colors, with the extra precision of float16.
 void applyCAMetalLayerColorSpace(void *nativeSurface,
                                  wgpu::TextureFormat format) {
+#if DEBUG
+  static dispatch_once_t threadCheck;
+  dispatch_once(&threadCheck, ^{
+    NSLog(@"[cametal-repro] CAMetalLayer main-thread check: %@",
+          NSThread.isMainThread ? @"PASS" : @"FAIL");
+  });
+  NSCAssert(NSThread.isMainThread,
+            @"CAMetalLayer mutations must run on the main thread");
+#endif
   CALayer *layer = (__bridge CALayer *)nativeSurface;
   if (![layer isKindOfClass:[CAMetalLayer class]]) {
     return;
@@ -30,14 +39,6 @@ void applyCAMetalLayerColorSpace(void *nativeSurface,
     // Restore the default (no color matching) when reconfiguring back to an
     // 8-bit format.
     metalLayer.colorspace = nil;
-  }
-  // The change must be set synchronously so the first present already sees
-  // it, and it must reach the render server. On a non-main thread (RN JS or
-  // worklet runtime) the property lands in that thread's implicit
-  // CATransaction, which may never commit on threads without a spinning
-  // runloop, so flush it now. On the main thread the runloop commits it.
-  if (!NSThread.isMainThread) {
-    [CATransaction flush];
   }
 }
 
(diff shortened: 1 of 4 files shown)


wcandillon:
Thank you for this, I need to think about it a bit more deeper

wcandillon:
@AlirezaHadjar just wanted to tell you that I am not dropping this issue. Currently looking at the same issue on Skia Graphite, once we fix there, I will circle back on this one. It just need a bit of more designing

PR #463, opened 2026-09-01
fix(💥): fix hot reload prototype cache 

### LINKED ISSUE #461: SIGSEGV in NativeObject<RNWebGPU>::installConstructor when install() runs after an expo-updates OTA apply (Android)
## Environment

- react-native-webgpu 0.5.15 (via the react-native-wgpu shim)
- react-native 0.85.3, Expo SDK 56, expo-updates ~56.0.21, Hermes
- HUAWEI MatePad 12 X (LRT-W09), Android 12 (API 31), arm64

## What happens

When an expo-updates OTA applies, the JS runtime is torn down and recreated in the same process. The app then crashes with a SIGSEGV null deref during WebGPU install. The relaunch is clean because the update is already installed, so it costs exactly one crash per OTA apply. Reproduced on every OTA apply on this device; we have not tried other hardware yet.

## Stack (from `dumpsys dropbox data_app_native_crash`)

```
#00 libjsi.so                 facebook::jsi::Value::~Value()+24
#01 libreact-native-webgpu.so jsi::Object::setProperty(...)
#02 libreact-native-webgpu.so rnwgpu::NativeObject<rnwgpu::RNWebGPU>::installConstructor+512
#03 libreact-native-webgpu.so rnwgpu::RNWebGPUManager::RNWebGPUManager(...)
#05 libreact-native-webgpu.so Java_com_webgpu_WebGPUModule_initializeNative
#08 base.vdex                 com.webgpu.WebGPUModule.install
```

## What I think happens

`install()` runs eagerly at JS module evaluation (`const _installOk = WebGPUModule.install()` in `src/main/index.tsx`), so it re-runs on the recreated runtime after the OTA apply. The NativeObject prototype cache is static and process-wide, and the runtime-aware swap compares runtime pointers. If the recreated runtime lands on the freed runtime's address, which is likely for an in-process recreate, `installConstructor` reuses stale `jsi::Object`s from the dead runtime and `setProperty` destructs a dead-runtime `Value`.

This looks like the reload state-isolation problem #427 addresses. In that thread you asked whether RN WebGPU still crashes on hot reload or OTA and how easily it reproduces: for us, 100% of OTA applies on this device. Happy to test a patch build.

wcandillon:
fixes #461

### DIFF
--- a/apps/example/src/App.tsx
+++ b/apps/example/src/App.tsx
@@ -38,6 +38,7 @@ import { AsyncStarvation } from "./Diagnostics/AsyncStarvation";
 import { DeviceLostHang } from "./Diagnostics/DeviceLostHang";
 import { DiagnosticsList } from "./Diagnostics/DiagnosticsList";
 import { WorkletRequestAdapter } from "./Diagnostics/WorkletRequestAdapter";
+import { ReloadLifecycle } from "./Diagnostics/ReloadLifecycle";
 import { ContextEdgeCases } from "./Diagnostics/ContextEdgeCases";
 import { ViewFormatsUseAfterFree } from "./Diagnostics/ViewFormatsUseAfterFree";
 import { RenderAfterUnmount } from "./Diagnostics/RenderAfterUnmount";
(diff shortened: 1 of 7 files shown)


wcandillon:
@kimchouard I technically couldn't reproduce the issue locally. If I were to give you a test package, is that something where you could confirm quickly if the issue is fixed or not?

Issue #468, opened 2026-09-02
Android: SIGSEGV in dawn::native::vulkan::FencedDeleter::DeleteWhenUnused when a Canvas unmounts (Surface::~Surface via onDropViewInstance)


eduardoborges:
**Version:** react-native-webgpu 0.9.0, three 0.184.0 (WebGPURenderer), React Native 0.87.1 (New Architecture, Hermes), react-native-reanimated 4.6.0 / react-native-worklets 0.12.1.

**Device:** Android emulator Pixel_10 (API 37, arm64, `-gpu host`, Vulkan `ranchu`), NDK 27.1, minSdk 26.

**What happens**

A screen with one `<Canvas transparent />` renders a glTF with three's `WebGPURenderer` (render loop via `setAnimationLoop`, `context.present()` each frame). Rendering, textures and gestures work. When the screen is popped (React Navigation native-stack) and the `Canvas` unmounts, the app dies with a native crash in roughly 2 out of 3 unmounts:

```
F libc    : Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x20
  dawn::native::vulkan::FencedDeleter::DeleteWhenUnused
  dawn::native::vulkan::SwapChain::DetachFromSurfaceImpl
  dawn::native::Surface::~Surface
  com.webgpu.WebGPUView.destroy
  com.webgpu.WebGPUViewManager.onDropViewInstance
```

The crash happens 0.1 s to 0.4 s after the React effect cleanup runs. In some runs it is preceded by `I WebGPU: GPU Device Lost (Destroyed): Device was destroyed.`, in others there is no device-lost line at all. iOS (Metal) never crashes with the same code.

**What I tried from JS, none of it changes the outcome**

1. `renderer.setAnimationLoop(null)` then `renderer.dispose()` in the effect cleanup (baseline): crash 3/3.
2. Same plus `context.unconfigure()` before dispose (confirmed it executes): crash 2/3.
3. Same plus keeping a strong reference to the renderer (and therefore the `GPUDevice`) for 3 s after unmount, so the device cannot be collected before the native view is dropped: crash 2/3, with no "Device Lost" line, so the device was alive when `Surface::~Surface` ran.

Timeline of one crash (logcat, `-v time`), with `console.warn` markers from JS:

```
14:40:52.887 W/ReactNativeJS: MedalShowcase: cleanup start
[...]

PR #469, opened 2026-09-02
fix(💥): detach swapchains before destroying the device (fixes Android Canvas unmount crash)

eduardoborges:
Fixes the native crash reported in #468.

## The bug

When a `<Canvas>` unmounts on Android, the app crashes about 2 out of 3 times with a SIGSEGV in Dawn's Vulkan backend:

```
F libc    : Fatal signal 11 (SIGSEGV), fault addr 0x20
  dawn::native::vulkan::FencedDeleter::DeleteWhenUnused
  dawn::native::vulkan::SwapChain::DetachFromSurfaceImpl
  dawn::native::Surface::~Surface
  rnwgpu::SurfaceInfo::detach(bool)
  Java_com_webgpu_WebGPUView_onViewDestroyed
```

iOS (Metal) never crashes with the same code.

## Root cause

three.js's `WebGPURenderer.dispose()` calls `device.destroy()` synchronously in the React effect cleanup. On Android the native view is dropped one frame later, so `SurfaceInfo::detach()` releases a `wgpu::Surface` whose swapchain is still attached. `~Surface` then runs `SwapChain::DetachFromSurfaceImpl`, which reads the already-destroyed device's `FencedDeleter`. That pointer is freed, so Dawn dereferences null at `0x20`.

Calling `Surface.Unconfigure()` at destroy time is not enough. What touches the device is the swapchain detach inside `~Surface`, so the surface has to be released (its destructor has to run) while the device is still alive.

## The fix

`GPUDevice::destroy()` now walks the `SurfaceRegistry` and, for every `SurfaceInfo` bound to this device, unconfigures and releases the `wgpu::Surface` right away, before `_instance.Destroy()`. Releasing it runs `~Surface` and `SwapChain::DetachFromSurface` against a live device. The native window pointer stays in `_nativeSurface`, so the later `SurfaceInfo::detach()` from the view teardown still returns the window to the platform.

Two small additions, with no change on the happy path. `SurfaceInfo::unconfigureIfDevice(device)` detaches and releases the surface when it belongs to `device`, and `SurfaceRegistry::unconfigureDevice(device)` calls it for every entry.

## Verification

Pixel emulator, Android API 37, `-gpu host` (real Vulkan via virtio-gpu), `minSdkVersion` 26, three.js `WebGPURenderer` rendering a glTF in a `<Canvas>` inside a React Navigation screen.

Before the fix: navigate in, wait for the first frame, press back, repeat. Crashes 2 out of 3 times.

After the fix: the same loop with `renderer.dispose()` called immediately in cleanup, no workaround on the app side. Five clean exits out of five, same PID throughout. Instrumentation confirms the `detach` after teardown now sees `surface=0`, already released.

It also runs on iOS, where `unconfigureDevice` is a cheap walk that does nothing but keeps both platforms on the same teardown path.

A standalone reproduction is at https://github.com/eduardoborges/react-native-webgpu-unmount-repro.

### DIFF
--- a/apps/example/src/Diagnostics/ContextEdgeCases.tsx
+++ b/apps/example/src/Diagnostics/ContextEdgeCases.tsx
@@ -1,4 +1,4 @@
-import React, { useRef } from "react";
+import React, { useRef, useState } from "react";
 import { Button, ScrollView, Text, View } from "react-native";
 import type { CanvasRef } from "react-native-webgpu";
 import { Canvas } from "react-native-webgpu";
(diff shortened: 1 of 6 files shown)


wcandillon:
Let me know if you can provide me with a reproducible example

eduardoborges:
Standalone reproduction (RN 0.87.1, react-native-webgpu 0.9.0, three 0.184.0): https://github.com/eduardoborges/react-native-webgpu-unmount-repro

Mount the medal, then hide it. On an Android emulator with a real Vulkan GPU (`-gpu host`) the app crashes on hide without this PR, and survives every toggle with it.

wcandillon (approved):
That's a great find, thank you for this 🙏

wcandillon:
@eduardoborges I've made a bunch of commits to your PR but I will revert it. I am considering a completely different approach to that problem

eduardoborges:
@wcandillon no problem! feel at home

wcandillon:
@eduardoborges Sorry I messed your branch and thank you for bringing this to my attention. I've made a PR against Dawn to fix it: https://github.com/google/dawn/pull/84
Now I am on the verge to merge/publish a workaround for it in the meantime: #470 

closing in favor of #470

PR #471, opened 2026-09-09
fix(⬆️): upgrade to chrome/154 and SPM support

wcandillon:


chrfalch:
Had a look — the distribution half looks right to me. A zipped xcframework with a published SHA-256, consumed by `binaryTarget(url:checksum:)` pointing straight at the release asset, and the manifest generated from a template so the URL and checksum can't drift by hand. No absolute paths in the committed project either. That's the pattern I'd want for Skia too.

The one thing I'd flag is this comment in `Package.swift`:

> React Native's CLI has no SPM autolinking for third-party TurboModule/Fabric libraries yet

That's no longer the case as of 0.87. A library that ships its own `Package.swift` is treated as *self-managed*: the autolinker references it through a symlink under the app's build dir and it consumes React Native's headers as products — `ReactHeaders`, `ReactNativeHeaders`, `ReactNativeDependenciesHeaders`. The codegen headers are covered too: they come from the generated `build/generated/ios` package as `ReactAppHeaders`, rather than from a per-project `ReactCodegen` pod. Shopify/react-native-skia#4043 does exactly this and builds Debug and Release with no CocoaPods at all, so if you'd rather not depend on `pod install` here, that road is open.

Three smaller things in the current approach:

1. **`PODS_ROOT` read at manifest-evaluation time is fragile, because SwiftPM caches manifests.** I hit this last week: a `fatalError` in a manifest didn't re-fire until I passed `--manifest-cache none`. Your comment notes the variable is unset while Xcode indexes — with caching, an evaluation from that moment can persist and leave the `-I` list empty, which surfaces as puzzling header-not-found errors rather than anything pointing at `PODS_ROOT`.
2. **`unsafeFlags` caps distribution.** SwiftPM only allows them for local/path dependencies, so this package can't be consumed by URL and version later. Fine for the local-package model, just worth knowing it's a one-way door.
3. **`Pods/Headers/Public` is the static-library layout.** The example's Podfile still honours `USE_FRAMEWORKS`, and in that mode those paths won't resolve.

One thing worth building in now, while you're in this file: Skia Graphite and WebGPU have to link the exact same Dawn. The podspec enforces that today by comparing `libs/.dawn-version` against this package's `dawn` field and refusing to install on a mismatch. SwiftPM has no `prepare_command`, so that check needs a new home — `Package.swift` is plain Swift and can read both files. Without it a mismatched pair links fine and crashes at runtime.

Happy to help with any of this.

Issue #293, opened 2025-12-17
Memory is not released


Nichel-777:
When exiting the WebGPU rendering page, the memory is not released.

<img width="1401" height="496" alt="Image" src="https://github.com/user-attachments/assets/7d8e5f86-b1a0-4332-bd2f-5f21c8936685" />

Nichel-777:
I’m developing on an Android device. This screenshot was taken from Android Studio. In fact, the issue I’m encountering is that when I repeatedly enter the rendering page, my graphics memory keeps increasing.

PR #478, opened 2026-09-15
fix(🐛): release the imported surface in GPUExternalTexture.destroy()

MiloszFilimowski:
## The problem

Importing a camera frame every frame with `importExternalTexture({ source })` and calling `texture.destroy()` + `frame.release()` right after `queue.submit()` still keeps the imported buffer alive until the JS garbage collector runs. At 30 fps on Android that is about 3.7 MB of graphics memory per 1280x720 frame, so memory climbs by ~110 MB/s and only comes back down when Hermes collects the wrappers, every few seconds. If nothing is submitted after the import (a frame processor that skips a frame), it never comes down at all.

This is with the usage from the `ImportExternalTexture` and `VisionCamera` examples: import, sample in a render pass, submit, `destroy()`, `release()`.

Probably related to #293.

## Root cause

`destroy()` calls `EndAccess` and drops the `wgpu::Texture` and `wgpu::SharedTextureMemory` handles, but keeps `_instance` (the `wgpu::ExternalTexture`). In Dawn, `ExternalTextureBase::DestroyImpl` only sets the state to `Destroyed` and keeps `mTextureViews`. The views hold the `SharedTexture`, and the `SharedTexture` holds the imported `VkImage` and `VkDeviceMemory` (`mSharedTextureMemoryObjects`) until `SharedTexture::DestroyImpl` runs. Any bind group that sampled the external texture holds a `Ref<ExternalTextureBase>` as well.

So the import lives as long as the JS wrappers for the external texture and its bind groups. Those wrappers report about 1 KB of memory pressure each, so a frame runtime that allocates little else does not collect for several seconds, and every frame in between adds another buffer.

Dawn releases fenced deletions from `Device::Tick`, which runs inside `queue.submit()`. That is why the "import but never submit" case grows without bound: nothing ever ticks.

## The fix

`destroy()` now calls `Destroy()` on the shared texture and on the external texture after `EndAccess`, and drops the source frame. Dawn then queues the `VkImage`/`VkDeviceMemory` (or the IOSurface texture on Apple) for deletion once the GPU work that used them completes, regardless of when the JS wrappers get collected. Both calls are safe right after the submit that sampled the texture for the same reason `EndAccess` is: Dawn keeps the resources alive for in-flight work.

On Android, `EndAccess` now chains `SharedTextureMemoryVkImageLayoutEndState`. Dawn's Vulkan backend rejects the call without it (`SharedTextureMemoryVk.cpp::EndAccessImpl` validates for exactly that branch), and `GPUSharedTextureMemory::endAccess` already does this. The Metal backend only accepts its own end state, so the chain stays Android-only.

`setLabel` checks for a destroyed instance.

## Numbers

Galaxy S23 (Android 14, Vulkan), react-native-webgpu 0.5.15, RN 0.86.3, release build, 1280x720 RGBA `AHardwareBuffer` at 30 fps. `dumpsys meminfo <pkg>` Graphics line sampled every 2 s.

| Per-frame work | Before | After |
|---|---|---|
| Camera only, no import | 76 MB flat | 76 MB flat |
| Import + one render pass + submit | 200 MB to 980 MB saw-tooth, drops every ~8 s (GC) | 145 MB flat |
| Import + `queue.submit([])`, nothing rendered | 220 MB to 2.6 GB saw-tooth, drops every ~20 s | 129 MB flat |
| Segmentation + blur, several passes | 300 MB to 1.1 GB saw-tooth | 175 to 190 MB flat over 3.5 min |
| After the loop stops | 620 MB held until the next GC | 118 MB |

30 fps with no dropped frames in every row, CPU a bit lower after the change (less GC work). On iOS (simulator, Metal) the footprint is flat before and after: the Metal texture references the IOSurface instead of allocating per import, so holding it only delays buffer reuse.

## Testing

- `yarn tsc` and `yarn lint` clean in `packages/webgpu`, cpplint clean with the CI flags.
- Built and run on the Galaxy S23 (release) and the iOS simulator with a camera effect pipeline (blur + segmentation, and a plain grayscale pass). Both platforms render correctly at 30 fps.
- New case in `ImportExternalTexture.spec.ts`: imports the same test frame three times, calls `destroy()` twice each time, and uses the label accessors on the destroyed texture. Skips on adapters without `rnwebgpu/native-texture`, like the existing test.

Repro for the leak, once per frame:

```ts
const frame = RNWebGPU.createVideoFrameFromNativeBuffer(nativeBuffer);
const texture = device.importExternalTexture({ source: frame });
// sample it in a render pass, queue.submit([...])
texture.destroy();
frame.release();
```

`adb shell dumpsys meminfo <package> | grep Graphics:` grows by ~110 MB/s before this change and stays flat after it.


wcandillon (approved):

wcandillon:
Thank you for this.

